PR #2, opened 2016-04-16
Add missing /dev/pts

tjstyle:
Missing /dev/pts causing "Can not write log, openpty() failed (/dev/pts not mounted?)"


XECDesign:
Thanks

PR #3, opened 2016-04-26
Correct github URL

lurch:


XECDesign:
That function is going to be changed to include the commit the image was built from anyway, so there is not much point merging this change if it will get overwritten soon anyway.

lurch:
Fair enough.

PR #4, opened 2016-05-02
Modifications to allow for basic Raspbian Lite build and image creation

spiderkeys:
Until there are official updates on creating the SD card and resolving a couple of issues, the following commits got me to a bootable Raspbian Lite image. This branch doesn't need to be merged, but I think creating a pull request will make the process visible for other people to work off until things are more cleanly tied up.

Couple of notes:
- Not sure why, but with my setup, the init_resize argument at the end of cmdline.txt sends my Pi into an endless reboot after it runs the resize script the first time. Looks like even without this cmdline option, the resize gets called by the normal init scripts:
```
[  OK  ] Started LSB: Resize the root filesystem to fill partition.
```
- Stage2 didn't end up with a populated fstab file, and boot failed as a result, so I uncommented the install step and created a slightly more elaborate default fstab
- Might want to look into using systemd-nspawn instead of chroot for host systems that are running systemd, to avoid potential issues with using systemd commands inside chroot's. I put a placeholder function, but haven't tested it since my current desktop doesn't use systemd
- I got a lot of your typical perl-related locale issues, so I configured the desired locale settings before installing the locales package to fix that.

One issue of note that I think should probably be addressed that I haven't sorted out yet is graceful failure. If you have a script that is set with bash -e and you experience a failure in on_chroot(), a lot of stuff gets left mounted and not cleaned up, which ends up breaking much of my userspace for some reason (can't spawn a new terminal, tab completions fail, tty's break, xserver will freeze if I switch between ttys, etc) and I end up having to restart my computer. Not sure what needs to be done here.

Looking forward to updates, hopefully some of this can be of use.


XECDesign:
Thanks, looks like there's a lot of good stuff here.

spiderkeys:
Closing this pull request due to several improvements included in recent commits. Opened new pull request with some updates I've made. See #5

PR #8, opened 2016-07-29
Fix for patchsets with same names in substages

staroselskii:
I used the provided *build.sh* script to generate a raspbian-lite image and came across an issue with network configuration.

If ```CLEAN``` variable is not set, image is not generated right. 

In that case (i.e. stage2/01-sys-tweaks and stage2/02-net-tweaks) a
directory with a same name (00-pc) is getting created for both quilt
patch sets. This might yield a situation when second patch set is not
applied.

The fix is to come up with a more sophisticated naming convention for intermediate
patchset directory (which .pc symlinks to) and create separate
directories for each patchset.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -42,8 +42,9 @@ EOF
 				rm -rf *-pc
 			fi
 			QUILT_PATCHES=${SUB_STAGE_DIR}/${i}-patches
-			mkdir -p ${i}-pc
-			ln -sf ${i}-pc .pc
+			SUB_STAGE_QUILT_PATCH_DIR="$(basename $SUB_STAGE_DIR)-pc"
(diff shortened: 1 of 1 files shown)


XECDesign:
Thanks, much appreciated.

staroselskii:
@XECDesign 

I was just wondering what'd been done in order to overcome the issue during the release cycle? A manual intervention had to be performed?

Or there's slight chance I'm the only one doing that without setting ```CLEAN=1```?

XECDesign:
@staroselskii, I am not entirely sure what happened there. Although `CLEAN=1` is generally used, especially for the final image release, I would still expect the issue to have come up at some point.

PR #9, opened 2016-09-01
Minor README.md updates

iKarith:
Removed TODO items for exporting image/NOOBS as they're done now, and added pxz to the list of requirements for the latter.

Willing to write some of the documentation to fill out README.md a bit as I learn how this tool works a bit more as I become more comfortable using pi-gen. This tool is a godsend for Raspple II which has been modifying the Raspbian NOOBS image after the fact with a manual process, so an automated tool we can alter and track changes will make our lives much easier.  ...Now if we could do the same for i686/amd64 Debian!  ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 #TODO
 
-1. Image export
-1. NOOBS export
 1. Simplify running a single stage
 1. Documentation
 
--- a/README.md
+++ b/README.md
@@ -7,4 +7,4 @@
 
 #Dependencies
 
-`quilt kpartx realpath qemu-user-static debootstrap zerofree`
+`quilt kpartx realpath qemu-user-static debootstrap zerofree pxz`


XECDesign:
Many thanks. Any help is appreciated.

iKarith:
Aside from removing a couple of uses of cat for pedantry purposes ;) 
I see nothing really useful to contribute to pi-gen aside from maybe 
writing down what I learn about how to run and alter it.  I'll keep a 
clean fork for that purpose.

I'll probably fork another copy under the Raspple II project and more 
heavily modify that, altering stages 3 and 4 to suit our needs.  
Actually, the whole thing is modular enough that if I wrap a few 
things in platform checks, I should be able to get it to put together 
an image file for a Debian system the same way that can be used on 
VMWare and friends.  Probably add support within the Raspple II 
version for package files specific to a platform and config 
specifying what platform that may be.

Like I said, you've got a VERY elegant system here.

Joseph

XECDesign:
Thanks. I think the original idea is fairly elegant and since then I've found that debian's official livecd build system uses a similar approach. However, I think if I could do it again, I would definitely focus more on a cleaner implementation, maybe using python. The stages themselves are easy to maintain and track, but build.sh is a mess.

PR #10, opened 2016-09-02
README.md: Working toward basic documentation

iKarith:
Documented the config file for IMG_NAME and APT_PROXY (the latter sparsely), and described the existing stages used to build Raspbian as they are now.  The latter is somewhat superfluous to pi-gen itself, but the big picture view of generating Raspbian makes the RTFS job for figuring out how stages work an easy task.

Stage anatomy will follow soonish.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,11 @@ use a different `IMG_NAME` to avoid confusion.
 
 You can also define `APT_PROXY` here if you need to.
 
-#Stages
+#Stage Anatomy
+
+
+
+#Raspbian Stages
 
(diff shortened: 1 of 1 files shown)


XECDesign:
Thanks. Looks good.

One issue is that it seems a little too conversational. If a sentence starts with 'obviously', it either doesn't need to be there or it's not actually that obvious. Maybe I am being nitpicky, so I am happy to merge as is.

iKarith:
I'll go back over it again.

XECDesign:
Many thanks

PR #14, opened 2016-09-06
build.sh: Support comments in package files

iKarith:
This patch allows the use of hash comments inside patch files.  It's a
little ugly, but it strips comments and collapses all whitespace down to
single space characters between package names.  It handles comments
anywhere in a line, as well.

Was unsure if \ continuation of the long sed line or the inclusion of a
couple of lines of comments explaining what the sed expressions are
doing would be appreciated, so didn't include them in this patch.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -16,7 +16,8 @@ EOF
 		fi
 		if [ -f ${i}-packages-nr ]; then
 			log "Begin ${SUB_STAGE_DIR}/${i}-packages-nr"
-			PACKAGES=`cat $i-packages-nr | tr '\n' ' '`
+			PACKAGES="$(sed -e ':a;N;$ !b a' -e 's/[[:space:]]*\(#[^\n]*\)*[[:space:]]/ /g' < ${i}-packages-nr)"
+
 			if [ -n "$PACKAGES" ]; then
 				on_chroot sh -e - << EOF
 apt-get install --no-install-recommends -y $PACKAGES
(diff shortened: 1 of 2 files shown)


XECDesign:
If it improves readability, \ continuation and comments are welcome. sed and regex generally look like a cat ran across the keyboard, so it could use some clarification. Without checking the manual, I have no idea what `:a;N;$ !b a` means.

iKarith:
I'll add comments and break the line.

Joseph
Sent via mobile

iKarith:
For the record, this is the file I'm testing with:

```
#comment
# comment with # in comment

# comment

ssh locales less fbset sudo psmisc strace module-init-tools ed ncdu crda
console-setup     #comment   
keyboard-configuration debconf-utils parted unzip
build-essential manpages-dev python bash-completion gdb pkg-config
python-rpi.gpio        v4l-utils
avahi-daemon


lua5.1
luajit #comment
hardlink ca-certificates curl
fake-hwclock ntp nfs-common usbutils
	libraspberrypi-dev libraspberrypi-doc libfreetype6-dev
dosfstools

# comment
dphys-swapfile
raspberrypi-sys-mods
pi-bluetooth
apt-listchanges
usb-modeswitch
```

Note the mix of spaces, tabs, lines with leading and trailing whitespace, etc.  If there's a condition I haven't considered, let me know and I'll test it before the PR gets merged.  Actually my previous sub pattern doesn't cope with this current test very well--what I'm about to commit does.

XECDesign:
Thanks

PR #15, opened 2016-09-10
Implement Landrash's suggestions for README.md

iKarith:


iKarith:
That shows a lot of commits which have already been merged, but were collapsed in the RPi-Distro tree into just a few.  I'm not sure how to make the collapses stay collapsed on my end.

XECDesign:
you can run git `rebase -i HEAD~n`, where n is the number of commits you want to squash. Then in the editor that comes up, you can squash or fix depending on what you actually want to do. `man git-rebase` might be worth a read. Luckily github has a squash feature, so I should be able to merge it as is.

iKarith:
I'd used rebase, but not rebase -i, thanks!

Joseph

Issue #13, opened 2016-09-05
Suggestion to build.sh


ssfdre38:
This script adds a check to make sure the user is under root to run the script and adds all dependencies before starting the scripting process and if IMG_NAME is not set, will echo how to set IMG_NAME in the command.
[build.txt](https://github.com/RPi-Distro/pi-gen/files/454448/build.txt)

ssfdre38:
This one i changed some stuff around and set it up where it will check for root access first, Makes sure the IMG_NAME is set, and updates the apt list and adds the dependencies before it starts so the system is set up just right for the build.
[build.txt](https://github.com/RPi-Distro/pi-gen/files/454471/build.txt)

XECDesign:
I don't know if it makes sense to check the dependencies every time and that apt-get line is a little expensive. Although I only support Mint, I don't want to add lines which I know won't work on non-apt based distros.

We could do something like this:
``` bash
for DEP in quilt kpartx realpath qemu-arm-static debootstrap zerofree pxz; do
  if ! command -v $DEP > /dev/null; then
    MISSING_DEPS="$DEP $MISSING_DEPS"
  fi
done

if [ -n "$MISSING_DEPS" ]; then
  echo "Could not find the following dependencies: $MISSING_DEPS"
  exit 1
fi
```
The problem with that is that the user won't know that qemu-arm-static comes from qemu-user-static, but it's better than just failing later down the line.

iKarith:
If you use `pkgname:cmd` format, you can split that in a for loop to test for commands' existence and return a list of necessary packages if not found.  That would allow you to `apt-get` only if the check for missing dependencies finds any.

Suggest `hash` instead of `command -v`.  In this instance it's probably not necessary, but it's a good habit since the former actually looks for the file, rather than assuming it is or isn't there.

The dependency list should probably also include `zip`.  Most people will already have that installed, y'know, unless they don't.  Discovering you forgot to install it on a new system after 74 minutes of watching all the stages finish wasn't amusing.  ;)

ssfdre38:
yea i was using raspbian lite that i built myself using the script and when i did my test overnight i saw that i needed zip at the very end. But yea there should be a check for the packages first before starting so that way the base system is set up for the build.

ssfdre38:
So i was trying some commands and right now i have it set up where it will check for dpkg and if it is there will not try and install it other if not there it will install if package is not install.

[...]

PR #16, opened 2016-09-14
Dependency checks + important bugfix

iKarith:
Sorry for the delay on this, especially given the embarrassing bug fixed in d1b1b77 which will cause a build to fail.

The rest of this push contains the dependency checking script discussed in #13.  I think the rest of the issues discussed in that ticket are misunderstandings of the invocation of pi-gen.


XECDesign:
Thanks again

Issue #19, opened 2016-09-16
export-noobs failure if you use /dev/mapper


iKarith:
I posted part of this in my run of pi-gen in #13, and I've noticed I always get the same result:

```
[05:13:41] Begin /home/tjcarter/Source/RPi-Distro/pi-gen/export-noobs
[05:13:41] Begin /home/tjcarter/Source/RPi-Distro/pi-gen/export-noobs/prerun.sh
umount: /: target is busy
        (In some cases useful info about processes that
         use the device is found by lsof(8) or fuser(1).)

real    72m34.969s
user    70m47.640s
sys     2m21.440s
```

Note this was my RPi-Distro tree, which has no patches.

You'll note that there is no `End /home/tjcarter/Source/RPi-Distro/pi-gen/export-noobs/prerun.sh` to match, and it ends with umount returning an error.  (An error in any stage causes pi-gen to bail, leaving filesystems mounted and other issues, but that's a slightly more complex issue.)  The prerun.sh is failing in its final line to `unmount_image ${IMG_FILE}`.  At first glance, `IMG_FILE` was unmounted, but a second look spots the `kpartx` call which would've placed the image back in use.

Again, this is a problem I see in RPi-Distro's tree, however I'm going to use my own tree to try and diagnose the problem because it has a completely populated work/ from today, resulting in a 15 minute build instead of a 70-90 minute build depending on disk caching.

My debug procedure mostly consists of running build.sh with the following patch:

```diff
diff --git a/export-noobs/prerun.sh b/export-noobs/prerun.sh
index a531be6..bde31b7 100755
--- a/export-noobs/prerun.sh
+++ b/export-noobs/prerun.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash -ex
 
 IMG_FILE="${STAGE_WORK_DIR}/${IMG_DATE}-${IMG_NAME}${IMG_SUFFIX}.img"
 NOOBS_DIR="${STAGE_WORK_DIR}/${IMG_DATE}-${IMG_NAME}${IMG_SUFFIX}"
```

Relevant output follows:

```
[00:35:45] End /home/tjcarter/Source/RasppleII/pi-gen/stage4
[00:35:45] Begin /home/tjcarter/Source/RasppleII/pi-gen/export-image
[00:35:45] Begin /home/tjcarter/Source/RasppleII/pi-gen/export-image/prerun.sh
[...]

PR #20, opened 2016-09-16
Cosmetic renaming

iKarith:
Seemed like it was a good idea to use the check_deps.sh file that was already there, and while I was at it (since there's also a .sed file in there) I figured I should also rename common to common.sh while I was there.

Also added `# bash #` to the top of each line since .sh isn't enough to tell an editor what kind of shell you're looking at.  Vim at least looks for bash or ksh on the first line to catch something in the shebang line.  Of course a sourced fragment hasn't got one, so we faked it.

Renamed the dependencies_check function to match the file.


iKarith:
I'd retitle this PR if I could because it now contains a fix for the unmount_image problems reported in #19 by virtue of me neglecting to put the cosmetics in a branch like I should have.

Opted for the --one-file-system option to tar rather than --exclude, which of course will have to be changed back if export-noobs is ever modified to operate on the stage's chroot rather than its generated image.  But it works now as written and no longer tries to unmount / if you use LUKS or LVM or something using /dev/mapper.

iKarith:
I'm going to close this, revert the cosmetic changes, and resubmit the bugfix as its own PR.

PR #21, opened 2016-09-29
Add MAX_STAGE and RUN_STAGE variables

gregmac:
* `MAX_STAGE`, to only run up to this stage. (default `4`, eg: `MAX_STAGE=2`)
 * `RUN_STAGE`, to only run a single stage (eg: `RUN_STAGE=1`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 #TODO
 
-1. Simplify running a single stage
 1. Documentation
 
 #Dependencies
@@ -17,6 +16,8 @@ The following environment variables are supported:
 
  * `IMG_NAME`, the name of the distribution to build (required)
  * `APT_PROXY`, proxy/cache URL to be included in the build
+ * `MAX_STAGE`, to only run up to this stage. (default `4`, eg: `MAX_STAGE=2`)
+ * `RUN_STAGE`, to only run a single stage (eg: `RUN_STAGE=1`)
 
 A simple example for building Raspbian:
 
--- a/build.sh
+++ b/build.sh
@@ -176,7 +176,7 @@ log "Begin ${BASE_DIR}"
 
 for STAGE_DIR in ${BASE_DIR}/stage*; do
 	STAGE_DIR_NUM=$(echo $STAGE_DIR | grep -o -E "[0-9]+$")
-	if [[ (-n $RUN_STAGE && $STAGE_DIR_NUM -eq $RUN_STAGE) && $STAGE_DIR_NUM -le $MAX_STAGE ]]; then
+	if [[ (-z $RUN_STAGE || $STAGE_DIR_NUM -eq $RUN_STAGE) && $STAGE_DIR_NUM -le $MAX_STAGE ]]; then
 		run_stage
 	else
 		echo "Skipping ${STAGE_DIR}"
--- a/build.sh
+++ b/build.sh
@@ -128,12 +128,9 @@ if [ -z "${IMG_NAME}" ]; then
 	exit 1
 fi
 
-if [ -z "${MAX_STAGE}" ]; then
-	MAX_STAGE=4
-fi
-if [ -n $RUN_STAGE ]; then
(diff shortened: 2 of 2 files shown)


Landrash:
I would suggest extending this a bit further. This could be used in a variety of ways such as building a set with specific features.
Excluding stages and/or specifying stages.

Example:
You only want to run stage 1-3 & 5-8
`EXCLUDE=4`

You want to run stage 1-3 & 8
`RUN_STAGE=1,2,3,8`

XECDesign:
Currently, I add a 'SKIP' file to the stages I don't want to run, but that's definitely not ideal.

I like Landrash's idea, but for it to make sense to run stages 1-3, skip 4 and run 5-8, I think the logic would need a bit of rework. Would you want to skip a stage in the middle because you don't want the changes from the previous stages to propagate or because you don't want that stage at all

 It would need to figure out which stage to copy the files from - the old stage 4 or the new stage 3? The logic would also need to do the right thing if you're only running 5-8 (which I think would be to copy the files from stage 4).

Landrash:
Didn't consider the problems with the logic @XECDesign 
This most likely goes out of scope for this but maybe having a dependency setting for each step.
Step 2 depends on step 1. Step 3 depends on step 2. Step 5 on the other hand only depends on step 3 so step 4 can be skipped. This would require some kind of sanity check before running on the other hand.

gregmac:
I saw these two settings as useful for two purposes:

* `MAX_STAGE` for doing a 'lite' build (eg, only up to stage2), which actually was my motivation 
* `RUN_STAGE` for debugging, which I really only did because it was in the README as a TODO and it was an easy change

I don't think skipping stages or building a complex dependency system is the appropriate way to build completely different images. For example, trying to build a single git repository that can build every possible Raspbian-derived distro that exists today would be crazy, so if you build it to support more than one, where do you draw the line?

I'd prefer to leave this PR with my original intent of debug usage. If you want to build different distros:

* Create branches -- you can still merge between them but changes stay separated
* We modify the build script and folder structure so you get, eg:
````
     mydistro/stage0
     mydistro/stage1
     mydistro/stage2
     mydistro-alt/stage0 -> symlink to ../config1/stage0
     mydistro-alt/stage1 -> symlink to ../config1/stage1
     mydistro-alt/stage2
     mydistro-alt/stage3
````
   Then you can run `DISTRO=mydistro-alt build.sh`
* Fork to a separate repository -- if they really are completely different distros

iKarith:
I'd suggest a default of 0 or unset, with either of those being a disable condition.

gregmac:
@iKarith Not sure what you mean. Right now, if you leave both params off, it runs identically to how it does without this change.

iKarith:
You said the default is 4.  I'm suggesting that there ought not be a default or that the default should be to do anything.  So if I create a stage 5 (as I've done for Raspple II for example), it ought to just do its thing until I start specifying minimal builds or something.

I've got some idea in mind I'm not 100% sure how to propose an implementation of related to this stuff, and the patch that exists is always worth 100 times the one that someone thinks they might be able to write.  Plus, this is an internal development tool even if it gets used externally, it's expected to change.

gregmac:
@iKarith Sorry I was incorrect, if not specified it really did default to stage4. Updated so it will run with any number of stages when not specified.

XECDesign:
It would also make sense to take out the 'SKIP' file check on line 91 to avoid redundant behaviour.

gregmac:
@XECDesign Done. I left in support for sub-stage SKIP files (line 103) because my changes don't provide the equivalent functionality for sub-stages, but I can go either way on it.

XECDesign:
Many thanks

XECDesign:
This was a bit broken after all.
run_stage sets PREV_ROOTFS_DIR
When RUN_STAGE is set, this isn't set for the skipped stages, making the next stage fail on copy_previous.

PR #24, opened 2016-10-05
README.md: Improve config variable documentation

iKarith:
Just added some more verbose documentation of the config variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,9 @@ The following environment variables are supported:
 
  * `APT_PROXY` (Default: unset)
 
-   If you require the use of an apt proxy, set it here.  This proxy will be
-   included in the created image file.
+   If you require the use of an apt proxy, set it here.  This proxy setting
+   will not be included in the image, making it safe to use an `apt-cacher` or
+   similar package for development.
 
  * `MAX_STAGE` (Default: unset)
 
--- a/README.md
+++ b/README.md
@@ -14,10 +14,29 @@ environment variables.
 
 The following environment variables are supported:
 
- * `IMG_NAME`, the name of the distribution to build (required)
- * `APT_PROXY`, proxy/cache URL to be included in the build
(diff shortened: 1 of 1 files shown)


XECDesign:
What do you mean by "this proxy will be included in the created image file"?
https://github.com/RPi-Distro/pi-gen/blob/dev/export-image/03-finalise/01-run.sh#L10

The idea is that you can use a proxy, but it's not kept in the resulting image..

iKarith:
I missed where the proxy was removed. All the more reason why we need docs. :)

I'll patch that when I'm in front of a terminal. I don't know if I can squash it since I've already pushed to the server, but if it'll let me, I'll do that.

Joseph
Sent via mobile

XECDesign:
GitHub now has a 'squash and merge' option, so it's not important.

XECDesign:
Many thanks

iKarith:
That's actually good because I wanted to use apt-cacher myself to build this, actually.  Just enough time is spent downloading files that it would be nice if I didn't have to.  I didn't use the APT_PROXY setting because I thought it would be saved to the image.  Now that I know it won't be, I'm free to use it.  :D

(Do we maybe want a patch that makes it possible to preserve the proxy setting via config variable?  It's a simple if [ -n ... ] to do that.

XECDesign:
There were a few times when raspbian.org was down when an image had to be prepared, so this options was added. We also use apt-cacher-ng to redirect the raspberrypi.org repo to an internal testing repo.

I don't see many situations where you'd want to release an image with a proxy set, but if it's a useful feature for you (I guess for testing), I'd merge it.

iKarith:
Not something I'd use personally, but if you're creating an in-house build that you'd like to preconfigure to use the in-house proxy, it could be wanted by someone.

XECDesign:
Given that the user base of pi-gen isn't that large, I wouldn't bother with it.

PR #25, opened 2016-10-09
rsync single file system

chzchzchz:
If proc is mounted in rootfs (e.g., via chroot), rsync will try copying kcore
when permitted to cross file system boundaries.


XECDesign:
Not sure why you would be chrooted at that stage, but erring on the side of caution makes sense. Thanks.

PR #29, opened 2016-11-14
adding dosfstools as a build dependency

sajal:
`build.sh` complained about `mkdosfs` not being found. Installing `dosfstools` makes it go away.


XECDesign:
Thanks

PR #40, opened 2017-01-16
Added Docker support

cfstras:
Hi,
I don't know whether this is of use for anyone, but I modified this image generator so one can use it with docker. I'm building some lot of custom images, and having this in docker really helps. It needs `--privileged` to work (because we need loop-mounts), but no devicemapper anymore.

- replaced necessity for devicemapper (was being used through kpartx) by using parted and
  losetup with offsets
- added Dockerfile
- added dependency for parted and grep
- added hints to README.md


XECDesign:
Thanks, I'll give this a go and report back. Do loopback offsets work as expected here? I used to do it that way, but seem to recall that using offsets like that means that most tools are not aware of the partition table. For example, mkdosfs would then create a filesystem that continues to the end of the file, past the end of the partition. I'd have to try it out to be sure though.

cfstras:
That's a valid concern. I'll verify that in my image, and report back. But I do remember that losetup has an offset _and_ a length parameter, so I can probably correct that.

cfstras:
Okay, you were right. It created a too-big filesystem.

New rebased commit should fix that.

XECDesign:
Ran into a minor issue. debconf popped up and did not seem to be able to read keyboard input. Normally this wouldn't be an issue, since debconf options should be pre-seeded. However, if this isn't properly interactive, that makes troubleshooting issues and adding/editing patches somewhat tricky.

scripts/common
```diff
@@ -53,11 +53,10 @@ unmount_image(){
 	local LOOP_DEV=$(losetup -j ${1} | cut -f1 -d':')
 	if [ -n "${LOOP_DEV}" ]; then
 		local MOUNTED_DIR=$(mount | grep $(basename ${LOOP_DEV}) | head -n 1 | cut -f 3 -d ' ')
-		if [ -n "${MOUNTED_DIR}" ]; then
+		if [ -n "${MOUNTED_DIR}" && "${MOUNTED_DIR}" != "/" ]; then
```
XECDesign: `[: missing ']'`

I think this should be `[ -n "${MOUNTED_DIR}" ] && [ "${MOUNTED_DIR}" != "/" ];`

cfstras:
Makes sense. I was wondering where that "missing ]" came from.

XECDesign:
I think that console setup patch can be removed entirely. Pushed a commit which should do that.

XECDesign:
Can you update an image after making changes, or does it need to be re-created?

cfstras:
Once you have changed something in the scripts, and rebuilt the image, you can either use `--volumes-from=pigen_work` on a second container to reuse the `/work`, `/deploy` and `/output` bindings; or you can use `docker run -it` instead of `docker run -i` and omit the script part behind `bash ...` to get an interactive shell.

Just being curious, are there any side-effects you can expect from re-running on the existing stages or is it designed to update and continue gracefully?

I'll test your changes and rebase.

cfstras:
I fixed the "git: command not found" coming from update_issue, fixed a bug related to unmounting and (in the process) also applied proper quoting to the functions in scripts/common.

let me know whether you want any changes to those, and I can squash if needed.

XECDesign:
Originally, the idea was that running a stage twice wouldn't cause any problems with that stage, but the changes wouldn't propagate from one stage to another. If a change was made in stage 2, it wouldn't be copied across to stage 3.

I think this functionality has been broken for a while, as the .pc directories are not properly tracked. I tend to use CLEAN=1 90% of the time.

Given that the docker build command is a little unwieldy, would it make sense to add a docker-build.sh?

I would prefer that the relevant options were put in the config file rather than passed as -e parameters. IMG_NAME shouldn't change in a given branch, so it just makes the command required to run the build unnecessarily long.

cfstras:
Okay, added a script and rebased. Not extensively tested, but should do the job.

XECDesign:
Still have a minor issue with `-e IMG_NAME=$IMG_NAME`. Don't want to encourage environemnt variables where the config file makes more sense.

cfstras:
Oh, sorry, I didn't find the mention in the readme before.

So the file will just be like this?
```
IMG_NAME=Raspbian
APT_PROXY=
CLEAN=0
```
Maybe you should add a template to make that clear...

XECDesign:
That's on the imaginary todo list.

cfstras:
updated and rebased. seems to work well!

XECDesign:
One more question. Why rsync to output rather than bind deploy?

cfstras:
I wasn't sure whether deploy only gets written to once or whether there's a lot of work happening in it.
Bind-mounts can cause problems if you do symlinks or hardlinks in them, especially if you're using Docker for Windows (like I am)

Feel free to change it if it only gets those files copied in.

XECDesign:
It only gets copied in or has the archives created there directly.

build-docker.sh
```diff
@@ -0,0 +1,73 @@
...
+	time $DOCKER run --rm --privileged \
+		--volumes-from="${CONTAINER_NAME}" --name "${CONTAINER_NAME}_cont" \
+		pi-gen \
+		bash -e -o pipefail -c "dpkg-reconfigure qemu-user-static &&
+	cd /pi-gen; ./build.sh;
+	rsync -avPc work/*/build.log deploy/*.zip /output/" &
+	wait
+else
+	trap "echo 'got CTRL+C... please wait 5s'; docker stop -t 5 ${CONTAINER_NAME}" SIGINT SIGTERM
+	$DOCKER run --name "${CONTAINER_NAME}" --privileged \
```
sam3d: Is there a reason this doesn't also `--rm` the docker container after runtime?

cfstras: In case of a failed build, you'll want the volumes (work and deploy) to hang around. You could manually name them, but using the existing container and `--volumes-from=` in the `continue=1` step is easier.

cfstras:
Ready from my side. Ping me if it needs squashing/rebasing.

XECDesign:
Thank you

sam3d:
Just wanna also say thank you for this PR, will make my life so much easier <3

Issue #45, opened 2017-02-07
Remove dependency on kernel support for binfmt


sam3d:
In order to build for docker, the underlying kernel has to have support for `binfmt`, which is required by `qemu-user-static`. On Docker for mac, this works fantastically because the kernel already has the required support. However, when attempting to run on a fresh Debian VPS many kernels don't have this support.

I'm not sure if there's an easy solve. Many on the internet have suggested to use [a version of qemu with the `QEMU_EXECVE` set so compiling for ARM doesn't escape out and cause the build to fail](https://github.com/resin-io/qemu). A [special docker image](https://github.com/show0k/docker-miniconda-armv7) has been created that uses this prebuilt special version of qemu but I'm not sure how much use this is in this context.

Attempting to install qemu-user-static produces the following output:

```
Setting up qemu-user-static (1:2.1+dfsg-12+deb8u6) ...
update-binfmts: warning: Couldn't load the binfmt_misc module.
update-binfmts: warning: Couldn't load the binfmt_misc module.
...
update-binfmts: warning: Couldn't load the binfmt_misc module.
Setting up quilt (0.63-3) ...
```

And some people on `circleci` have documented [similar behaviour](https://discuss.circleci.com/t/add-binfmt-misc-support-for-cross-compilation/2264/3) as a failed result of cross-compilation for ARM architectures.

The actual error with running `IMG_NAME=example ./build.sh` occurs during the initial file extractions:

```
...
I: Extracting zlib1g...
I: Running command: chroot /root/pi-gen/work/2017-02-07-tiny/stage0/rootfs /debootstrap/debootstrap --second-stage
chroot: failed to run command '/debootstrap/debootstrap': Exec format error
```

While this isn't an issue inherent to `pi-gen`, it would be good to remove the dependency on underlying kernel support that isn't always present.

sam3d:
I think I've found a temporary workaround to this issue.

On the host machine (outside the docker container), you need to simply install `binfmt-support` there. It cannot be done from within the container (no matter the privileged mode). Not quite sure how to get around this issue though, as it appears to be a requirement.

XECDesign:
Not sure that this is an issue rather than just a limitation. Pi-gen only really supports whatever OS I happen to be running at the time and anything outside of that is too much headache, due to subtle differences like this.

sam3d:
[...]

PR #46, opened 2017-02-07
Document warning for binfmt-support requirement

sam3d:
Refs #45.

A bit of a weird one, as it'll only occur when attempting to build from `Docker`. As mentioned in the issue and the commit, if the underlying kernel doesn't have support for `binfmt` then the `qemu-user-static` installation will fail (pretty much) silently upon building the docker image.

The ensuing docker build process will then not be able to virtually chroot (and will also fail).

There is no easy fix for this as right now we need `qemu-user-static` for ARM emulation, so I've added a small bit of documentation in the README that'll hopefully save someone the hours of debugging I just went through!

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,12 @@ You can then remove the build container with `docker rm pigen_work`
 
 If something breaks along the line, you can edit the corresponding scripts, and
 continue:
+
 ```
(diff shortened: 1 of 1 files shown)


XECDesign:
Thanks

PR #52, opened 2017-03-01
Enable pi-gen build when "pi-gen/work" dir is a symlink

dcwangmit01:
* Occasionally, it is necessary to symbolically link the pi-gen/work
  directory to a different place on the filesystem.  This is true if
  the current partition is low on space, or when building within
  vagrant virtual machine shared mounts which do not support all
  regular fs operations.
* Without this change, the pi-gen build will fail to unmount because
  the OS 'mount' command returns canonical paths which fails to match
  the symbolically linked path to 'work' dir.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -132,7 +132,7 @@ export IMG_DATE=${IMG_DATE:-"$(date -u +%Y-%m-%d)"}
 
 export BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 export SCRIPT_DIR="${BASE_DIR}/scripts"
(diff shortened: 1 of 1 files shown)


XECDesign:
I wonder if it would make more sense to make WORK_DIR configurable through the config file. I'll merge it for now, but the behaviour may change in the future.

XECDesign:
Reverted (see https://github.com/RPi-Distro/pi-gen/issues/54)

dcwangmit01:
Apologies @XECDesign .  The PR issue was that readlink "-f" needed to be replaced to the "-m" where the path segments did not need to exist.

Sounds like the WORK_DIR variable in the config file is your preference.  I'll try that approach and submit a PR.

PR #56, opened 2017-03-17
Updated {i}-run-chroot script filename to match {i}-run.sh format

m-roberts:


XECDesign:
Thanks. I am guessing that since it took that long to notice, it's probably not a very useful feature and should be taken out completely.

m-roberts:
I am making use of this functionality in a private fork (with pi-gen as a git submodule), so it would be great if it could stay to ensure compatibility and simplicity on my end. I don't mind leaving it as-is - I only suggested this to standardise filenames - but I would prefer not to have it removed!

XECDesign:
Alright, thanks for the feedback

PR #57, opened 2017-03-17
Shorten config.txt url

bennuttall:
Because why not? :)

### DIFF
--- a/stage1/00-boot-files/files/config.txt
+++ b/stage1/00-boot-files/files/config.txt
@@ -1,5 +1,5 @@
 # For more options and information see
-# http://rpf.io/configtxtreadme
+# http://rpf.io/configtxt
 # Some settings may impact device functionality. See link above for details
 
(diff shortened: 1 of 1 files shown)


ZanderBrown:
Backwards compatibility?

XECDesign:
Might be a bit too late. That was the link I was given to use and raspberrypi-sys-mods updates it on existing installs:
https://github.com/RPi-Distro/raspberrypi-sys-mods/blob/master/debian/raspberrypi-sys-mods.postinst#L61

Now it would need to be something like:
If updating from this version or older, change this URL to this other URL.
If updating from this other version, change this other URL to yet another URL.

It would be nice to avoid touching config files whenever possible.
Do you feel like it's  enough of an issue?

bennuttall:
Not essential, just a shame the short url was created unnecessarily long.

XECDesign:
I think I'll merge this and sort out sys-mods next time it's modified. Thanks.

PR #68, opened 2017-05-06
README.md: Describe more config variables, cleanup

iKarith:
Added some documentation of environment variables to README.md,
specifically WORK_DIR and DEPLOY_DIR.  Explicitly did not include
SCRIPT_DIR at this time because that appears to require some small work
before it can be changed.

Took the chance for some other syntax cleanups while I was at it, mostly
breaking long lines and inserting whitespace where some Markdown parsers
require it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,24 @@
 # pi-gen
+
 _Tool used to create the raspberrypi.org Raspbian images_
+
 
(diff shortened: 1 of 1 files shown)


XECDesign:
Thanks

PR #69, opened 2017-05-17
Update stage specification example

jsiverskog:
The previous example failed for me due to a missing SKIP in
stage5. While at it, complete the example with all steps needed to get
an image generated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,8 +156,10 @@ from `./stage2` (if building a minimal system).
 
 ```bash
 # Example for building a lite system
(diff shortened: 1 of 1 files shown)


XECDesign:
Thanks, will merge after the next image release.

thomas-phillips-nz:
Should we also add some documentation about what Stage 5 does at the same time?

XECDesign:
Thanks. I'll add a description of Stage 5.

Issue #62, opened 2017-04-17
Repeated use of build-docker.sh eats disk space extremely fast


malaire:
Each use of `build-docker.sh` leaves behind tens of GBs of files in `/var/lib/docker` which are never deleted, not even with `docker rm pigen_work`.

This eats up disk space extremely fast, wasting easily hundreds of GB, with no way to get rid of these files (except `sudo rm -rf` which might destroy integrity of docker).

andrewmichaelsmith:
@malaire I think you want to do:

`docker rm -v pigen_work`

XECDesign:
Fixed by https://github.com/RPi-Distro/pi-gen/pull/70

PR #70, opened 2017-05-17
Tell docker to delete the volume as well

andrewmichaelsmith:
Addresses #62

Because of the use of volumes the .img files (and everything else) will stay around on disk even when you do `docker rm pigen_work`. Passing `-v` deletes the volume as well and frees up disk space.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ vi config         # Edit your config file. See above.
 ```
 
 If everything goes well, your finished image will be in the `deploy/` folder.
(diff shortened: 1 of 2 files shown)


jsiverskog:
Should probably update build-docker.sh as well ('echo "  docker rm $CONTAINER_NAME"').

andrewmichaelsmith:
@jsiverskog good spot - have updated the pull request. Thanks.

XECDesign:
Thanks, will merge after the next image release.

Issue #75, opened 2017-06-02
Paths with spaces do not work


thomas-phillips-nz:
Upon inspecting the code and testing the build system, there are many many places in several scripts where if the current path (pwd) has spaces in it, the script will fail.
These should be cleaned up so we don't have to rely on the pwd having no spaces.

I have seen talk of changing the build script to python, but I have seen the dependency check script fail too. Looks like the whole build system needs to be modified to handle such paths.

davesteele:
ff2d5ed may have helped.

XECDesign:
Just had a look at this and even if I start fixing the bits which are broken in pi-gen, debootstrap still struggles, at least in ubuntu 19.10 - https://bugs.launchpad.net/ubuntu/+source/debootstrap/+bug/1716079

PR #76, opened 2017-06-02
Fixed running docker script from within pwd containing spaces

thomas-phillips-nz:
I used the default `/dev/null:/dev/null` mount so the strings passed to the `docker run` invocation could stay tidy. This should have no impact on the process.

I also added `-e IMG_NAME=${IMG_NAME}` to allow `IMG_NAME` to be passed on the command line when invoking the script, since this one variable is usually the single variable that gets changed. This is partly because the script checks that the variable is set, but not where it comes from (if you specify the variable on the command line, it gets caught a lot later inside the docker container)
This also allows basic running of the image builder without needing to create a config file.

This pull request partly addresses #75, at least for the docker build.

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -12,9 +12,9 @@ if ! $DOCKER ps >/dev/null; then
 fi
 set -e
 
-config_mount=
+config_mount="/dev/null:/dev/null"
 if [ -f config ]; then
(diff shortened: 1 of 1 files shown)


XECDesign:
Thanks

Issue #72, opened 2017-05-24
The build date is always UTC, even with different time zone settings


g-e-r:
This took me some time to find out.
 My timezone - PDT - is UTC-7
 So if I start a build and it fails after 5pm, the date will change...
 That happens very often in the US West coast.
 there should be a better strategy for this.

thomas-phillips-nz:
In build.sh, line 131, it assigns the `IMG_DATE` variable to the current UTC date _if the IMG_DATE variable is not set_. You should be able to invoke the script using IMG_DATE=xxx ./build.sh if you want to set your own date.

That said, I don't like the fact that the date is UTC either. I propose we change it to always use local system time. It can always be overridden using the IMG_DATE variable.

XECDesign:
Fixed by https://github.com/RPi-Distro/pi-gen/pull/77

PR #77, opened 2017-06-02
Image build date now honours timezone

thomas-phillips-nz:
This is to address #72 by removing the UTC switch from the call to `date`. The docker build will still by default use UTC unless its timezone has been changed.

The variable `IMG_DATE` can be specified in the `config` file, to override this date too.


XECDesign:
Thanks

Issue #84, opened 2017-07-13
Error on build: "Invalid filesystem option set: ^metadata_csum,^huge_file" with docker build


guysoft:
Hey,
I am trying to build pi-gen with the docker build and without it I get the following error. There is no image at the end in the deploy folder.
Using Ubuntu 17.04:

```
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0xb2a851a7  

Device                                                                      Boot 
Start     End Sectors  Size Id Type
/pi-gen/work/2017-07-13-Raspbian/export-image/2017-07-13-Raspbian-lite.img1      
 8192   94207   86016   42M  c W95 FAT32 (LBA)
/pi-gen/work/2017-07-13-Raspbian/export-image/2017-07-13-Raspbian-lite.img2      
94208 3241835 3147628  1.5G 83 Linux


Command (m for help): The partition table has been altered.
Syncing disks.

/boot: offset 4194304, length 44040192
/:     offset 48234496, length 1611585536
mkfs.fat: warning - lowercase labels might not work properly with DOS or Windows
mke2fs 1.42.12 (29-Aug-2014) 
Invalid filesystem option set: ^metadata_csum,^huge_file
Done! Your image(s) should be in deploy/
/home/ubuntu
/var/lib/docker/volumes/66fb28dda4b5110681d778b8eacbb516798772ed703b99341caa9334b
de50762/_data/2017-07-13-Raspbian/export-image
```

current workaround is to build without docker

XECDesign:
You'll need have to revert that commit or use newer e2fsprogs.

```bash
$ truncate -s 1G test.img
$ mkfs.ext4 -O ^metadata_csum test.img 
mke2fs 1.42.13 (17-May-2015)
Invalid filesystem option set: ^metadata_csum
```
```bash
$ mkfs.ext4 -O ^metadata_csum test.img
mke2fs 1.43.4 (31-Jan-2017)
Discarding device blocks: done                            
Creating filesystem with 262144 4k blocks and 65536 inodes
Filesystem UUID: 1e400e09-73fb-4eba-b3f7-79cfa48303ae
Superblock backups stored on blocks: 
        32768, 98304, 163840, 229376

Allocating group tables: done                            
Writing inode tables: done                            
Creating journal (8192 blocks): done
Writing superblocks and filesystem accounting information: done
```

guysoft:
[...]

PR #85, opened 2017-07-16
Export-image: Only disable metadata_csum feature if it's possibly ena…

rkubes:
…bled.
This is to address #84 by checking the configuration file for mke2fs on the host system to search for "metadata_csum" before passing it as an option to mkfs.ext4.
The assumption is that if this text does not exist in the configuration file, then either the e2fsprogs is outdated and won't support it, or the feature is just not enabled on any of the configured defaults. Either way, the build script won't need to try to disable it when exporting the image.


cfstras:
Please don't change the behaviour of a distribution generator depending on the versions build system...

Either make it a switch to select in build.sh, or document the dependency on newer e2fsprogs.

XECDesign:
Thanks

PR #86, opened 2017-07-17
Added docker-compose.yml for easy apt-cacher-ng startup

cfstras:
Tested using docker-build.sh. Gives a huge speedup and makes it dead-simple to start a local APT proxy.


XECDesign:
Thanks again

PR #95, opened 2017-08-20
Check for curl as a dependency for export-image

rkubes:
This pull request is a result is to address a new dependency on "curl" that is currently unchecked.

Without this commit, the build process will get all the way to the export-image step, then fail if "curl" is not installed on the host system.

This dependency was added with:
7149e20f2d919efdafc325602c373aee35b27023 "export-image: generate .info file"
However, curl was not added to the depends file to ensure the dependency is satisfied.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ RUN apt-get -y update && \
     apt-get -y install \
         git vim parted \
         quilt realpath qemu-user-static debootstrap zerofree pxz zip dosfstools \
-        bsdtar libcap2-bin rsync grep udev xz-utils \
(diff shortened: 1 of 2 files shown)


rkubes:
I don't use the docker build and have no experience with docker. That said, it seemed simple enough. 

Can you verify that the additional commit will satisfy the requirement to add curl to the Docker build as well?

If it was more complex I would have left it to the team that maintains the docker build.

jacen92:
Your last commit is correct for docker :)
Thanks

XECDesign:
Thanks

PR #96, opened 2017-08-24
export-image: when work in a non-English locale environment the PARTUUID will silently fail, resulting to an unbootable image file. 

jamesruan:
Because in locale like zh_CN.utf-8 the `Disk identifier` becomes `磁盘标识符`. Sed will silently fail.
so I use dd and xxd to directly read the disk identifier out.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ RUN apt-get -y update && \
     apt-get -y install \
         git vim parted \
         quilt realpath qemu-user-static debootstrap zerofree pxz zip dosfstools \
-        bsdtar libcap2-bin rsync grep udev xz-utils curl \
+        bsdtar libcap2-bin rsync grep udev xz-utils curl xxd \
     && rm -rf /var/lib/apt/lists/*
 
 COPY . /pi-gen/
--- a/depends
+++ b/depends
@@ -12,3 +12,4 @@ grep
 rsync
 xz:xz-utils
 curl
+xxd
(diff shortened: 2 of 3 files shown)


XECDesign:
In that case xxd would also need to be added as a dependency.

jamesruan:
@XECDesign Is this OK? Simply add xdd to 'depends' file.

XECDesign:
Thanks

lurch:
Should this same logic also be used in https://github.com/RPi-Distro/raspi-config/blob/master/init_resize.sh ? I remember that if you change the language used by NOOBS, then Raspbian also automatically picks up that locale change? (or perhaps `init_resize.sh` gets run before the locale gets changed, even when using NOOBS?)

Also, only a minor point but the version of `xxd` in Ubuntu 14.04 doesn't have the `-e` flag used here.
```
$ xxd -h
Usage:
       xxd [options] [infile [outfile]]
    or
       xxd -r [-s [-]offset] [-c cols] [-ps] [infile [outfile]]
Options:
    -a          toggle autoskip: A single '*' replaces nul-lines. Default off.
    -b          binary digit dump (incompatible with -ps,-i,-r). Default hex.
    -c cols     format <cols> octets per line. Default 16 (-i: 12, -ps: 30).
    -E          show characters in EBCDIC. Default ASCII.
    -g          number of octets per group in normal output. Default 2.
    -h          print this summary.
    -i          output in C include file style.
    -l len      stop after <len> octets.
    -ps         output in postscript plain hexdump style.
    -r          reverse operation: convert (or patch) hexdump into binary.
    -r -s off   revert with <off> added to file positions found in hexdump.
    -s [+][-]seek  start at <seek> bytes abs. (or +: rel.) infile offset.
    -u          use upper case hex letters.
    -v          show version: "xxd V1.10 27oct98 by Juergen Weigert".
```

XECDesign:
@lurch Good question. I don't think locales would be set in that environment, but don't know off the top of my head. Happy to leave it as is until somebody verifies it's a problem.

PR #101, opened 2017-08-27
Add a Qemu mode

jacen92:
This branch contains some minor fixes and a new feature.

It will be possible to create Qemu compatible image (usefull to test fixes or changes in Qemu without having to setup a new real rpi).

This is possible with a parameter QEMU_MODE="true"|"false".
If QEMU_MODE == "true" the images are configured to boot on Qemu but not on a real RPI, -qemu is append at the end of the image file to avoid mistakes.

### DIFF
--- a/export-image/04-finalise/01-run.sh
+++ b/export-image/04-finalise/01-run.sh
@@ -15,8 +15,10 @@ fi
 rm -f ${ROOTFS_DIR}/etc/apt/apt.conf.d/51cache
 rm -f ${ROOTFS_DIR}/usr/sbin/policy-rc.d
 rm -f ${ROOTFS_DIR}/usr/bin/qemu-arm-static
-if [ -e ${ROOTFS_DIR}/etc/ld.so.preload.disabled ]; then
-        mv ${ROOTFS_DIR}/etc/ld.so.preload.disabled ${ROOTFS_DIR}/etc/ld.so.preload
+if [ "${USE_QEMU}" != "1" ]; then
+	if [ -e ${ROOTFS_DIR}/etc/ld.so.preload.disabled ]; then
+		mv ${ROOTFS_DIR}/etc/ld.so.preload.disabled ${ROOTFS_DIR}/etc/ld.so.preload
+	fi
 fi
 
 rm -f ${ROOTFS_DIR}/etc/apt/sources.list~
--- a/stage2/EXPORT_IMAGE
+++ b/stage2/EXPORT_IMAGE
@@ -1,1 +1,4 @@
 IMG_SUFFIX="-lite"
+if [ "${USE_QEMU}" = "1" ]; then
+	export IMG_SUFFIX="${IMG_SUFFIX}-qemu"
+fi
--- a/README.md
+++ b/README.md
@@ -66,6 +66,10 @@ The following environment variables are supported:
 
    Output directory for target system images and NOOBS bundles.
 
+ * `USE_QEMU` (Default: `"0"`)
+
+   This enable the Qemu mode and set filesystem and image suffix if set to 1.
+
 
 A simple example for building Raspbian:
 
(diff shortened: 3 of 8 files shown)


XECDesign:
I like the idea, but I think this will need to wait until the python rewrite to also address the modular builds issue.

jacen92:
Hi, what did you mean about python rewrite?
All images created with this feature will works. It does not care about
current stage.

XECDesign:
The plan is to rewrite pi-gen in python with a more flexible and sensible build flow.

I can merge this for now, since it's not a major change. Could you please rebase it and remove all the changes which are not relevant to the qemu build (formatting, for example).

jacen92:
Done

Thanks

jacen92:
If you want to tests it you can follow informations from https://github.com/jacen92/pi-gen/wiki/QEMU

Tell me if you want more informations

jacen92:
Rebased and ready to be merged.

export-image/04-finalise/01-run.sh
```diff
@@ -15,8 +15,10 @@ fi
 rm -f ${ROOTFS_DIR}/etc/apt/apt.conf.d/51cache
 rm -f ${ROOTFS_DIR}/usr/sbin/policy-rc.d
 rm -f ${ROOTFS_DIR}/usr/bin/qemu-arm-static
-if [ -e ${ROOTFS_DIR}/etc/ld.so.preload.disabled ]; then
-        mv ${ROOTFS_DIR}/etc/ld.so.preload.disabled ${ROOTFS_DIR}/etc/ld.so.preload
+if [ "$USE_QEMU" = "false" ]; then
```
m-roberts: Shouldn't this be a double equals?

XECDesign: Should it? Everything I've seen suggests that = and == mean the same thing (string comparison).

jacen92: And I undestand this like:
If we are not in qemu mode then proceed as usual

jacen92: In bash it works with just an single equal

jcwren: The bash reference manual says '=' and '==' are synonymous.

http://tldp.org/LDP/abs/html/comparison-ops.html

XECDesign: In fact, trying a `==` in dash seems to error out, so it looks like `=` is more portable, but it doesn't matter here.

jacen92: indeed, this feature is disabled by default.

XECDesign:
The only 'problem' I have is that it goes against convention of how everything else works in terms of how things are set. Elsewhere, the tests look more like [ "$USE_QEMU" = "1"  ] rather than true/false and no default would need to be set. It doesn't make sense to have  `CLEAN=1 USE_QEMU=true ./build.sh`

jacen92:
Ok I understand, I will fix this if you want.

XECDesign:
Yup, please.

jacen92:
Ok so build and tests in progress

Stay tuned

stage2/EXPORT_IMAGE
```diff
@@ -1 +1,4 @@
 IMG_SUFFIX="-lite"
+if [ $QEMU_MODE = "true" ]; then
```
XECDesign: Quote marks around `$QEMU_MODE` for cases when it's not set at all

jacen92: I am testing something like:

> 	if [ $USE_QEMU -ne 1 ]; then

XECDesign: Same problem. If `$USE_QEMU` is not set, then you get `bash: [: -ne: unary operator expected`

jacen92:
So I've tested and it works with qemu and a real board.

jacen92:
Could you please merge this?

XECDesign:
Yes, got a bit side-tracked.

Thanks

PR #103, opened 2017-08-31
[fixing] docker parts

jacen92:
Use $DOCKER in build-docker instead of just docker command when necessary and remove container if build succeed


XECDesign:
Will leave this open for a while for people who use docker to comment, in case there is something wrong.

jacen92:
I tried an hour ago and it works but I understand.

wait and see

cfstras:
works for me.

XECDesign:
Thank you both.

Issue #93, opened 2017-08-17
Feature Request: Modular Builds


duerig:
I've been working on a build script for my own images, similar to the user in #50. While deleting and replacing later stages works to an extent, it is hard to maintain, especially when pulling in new changes from the main repository. But I found that this could be made much easier with just a few minor changes to build.sh:

(1) Make a STAGE_DIRS variable which allows a space-separated list of stages instead of just using stage*. It would of course default to stage*. This would allow people who wanted custom stages to create separate directories for them and so not interfere with the default stage*. It would avoid building unwanted stages and make merging changes from upstream easier.

(2) Change the prerun.sh files to use a new 'copy_stage' that explicitly names the previous stage rather than 'copy_previous'. This would let users just rebuild later stages incrementally when doing development without needing to re-run earlier stages. copy_previous only works  by remembering the stages as they are run. This is the only piece of state that is kept from stage to stage so remove this one thing and incremental builds are easy.

With these two minor changes, you can make it very easy for somebody to check out the repo, create a 'stage-foobar' directory, add that and it's dependant stages to the config file, and start building a new image based on any one of the canonical stages. Then when they want to update their repo later to accept changes made here, it is a simple pull request that requires no deconflicting. And after the first run, they can tweak their config file to only re-run their own stage in future builds while they develop it to improve turnaround time for debugging.

I've got this working on my tree. If you are interested in doing this, I am happy to send a diff or pull request.

XECDesign:
Sounds like an improvement. A PR would be welcome.

jacen92:
Hello,,

Could we imagine a minimal desktop image? It should simplify the feature if we just have to set the last stage to build.

RichardBronosky:
"rebuild later stages incrementally when doing development without needing to re-run earlier stages" I'm trying to do this right now. Any suggestions on how to do this would be appreciated. How are you doing this now?

duerig:
I'll try to get my PR ready tonight. Sorry for not following up earlier.
-Jonathon Duerig


On Wed, Oct 4, 2017, at 07:34 AM, Bruno Bronosky wrote:
> "rebuild later stages incrementally when doing development without
> needing to re-run earlier stages" I'm trying to do this right now. Any
> suggestions on how to do this would be appreciated. How are you doing
> this now?> — You are receiving this because you authored the thread. Reply to
> this email directly, view it on GitHub[1], or mute the thread[2].> 


Links:

[...]

PR #109, opened 2017-09-12
Selection for the last stage to build

jacen92:
Should solve Feature Request: Modular Builds #93

Allows dev to set the last build step without adding SKIP file and removing stuff in the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,10 @@ The following environment variables are supported:
 
    Output directory for target system images and NOOBS bundles.
 
+ * `LAST_STAGE` (Default: `"5"`)
+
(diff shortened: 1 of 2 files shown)


XECDesign:
pi-gen should be fairly agnostic about what it's actually building. It shouldn't assume how many stages there are or which of them are exported. The validation check and default value don't quite make sense in that context.

jacen92:
but each step depends of the previous right? in fact they have an order an devs know which is the last if he add more stage. It is not dynamically set or changed

XECDesign:
Yes, but a person customising a distro or creating their own shouldn't need to touch build.sh and the documentation shouldn't assume we're building the official rpf raspbian image.

jacen92:
I will make a new one later

PR #112, opened 2017-09-21
Docker-Build enhancements

cfstras:
- don't use any volume mounts anymore to allow using docker-in-docker
(with docker.sock mounted)
- smaller Docker build context by ignoring some files


XECDesign:
Thanks

PR #113, opened 2017-09-21
fix docker-in-docker copying when not doing CONTINUE=1

cfstras:
Sorry for the second pull request -- when running this again I noticed it was deleting the container too early.

This fixes that and also deletes the container always -- only if the build was successful.


XECDesign:
Thanks

PR #135, opened 2017-11-22
Remove /etc/vnc/updateid when finalising

sconemad:
The file /etc/vnc/updateid contains an ID generated upon installation, which is supposed to be a unique ID used by the vnc update notification service. However, since this file is present on the disk image, it is the same for all installations.

This change removes this file when finalising the disk image, so it will get regenerated properly if/when the server is first run.


XECDesign:
Thanks

PR #136, opened 2017-12-14
Update README with some information on how the build works

bjpirt:
Having just spent some time making my own modifications to pi-gen to build my own images and having to learn how the build process works I thought it would be helpful to others if it were documented in the README so they can get going quicker. Also, less support questions...

Let me know if you'd like to see any modifications, also if I've got the details correct!

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,13 +3,13 @@
 _Tool used to create the raspberrypi.org Raspbian images_
 
 
-### TODO
-1. Documentation
-
-
 ## Dependencies
 
-On Debian-based systems:
+pi-gen runs on Debian based operating systems. Currently it is only supported on
+either Debian Stretch or Ubuntu Xenial and is known to have issues building on
+earlier releases of these systems.
+
+To install the required dependencies for pi-gen you should run:
 
 ```bash
 apt-get install quilt parted realpath qemu-user-static debootstrap zerofree pxz zip \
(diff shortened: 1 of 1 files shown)


XECDesign:
Looks good. Thanks.

README.md
```diff
@@ -179,3 +221,21 @@ sudo ./build.sh  # or ./build-docker.sh
...
+
+## Skipping stages to speed up development
+
+If you're working on a specific stage the recommended development process is as
+follows:
+
+ * Remove the EXPORT_IMAGE and EXPORT_NOOBS files until you're ready to actually
+   generate an image
+ * Add SKIP files to the stages you don't want to build. For example, if you're
+   basing your image on the lite image you would add these to stages 3, 4 and 5.
```
jcwren: I need to re-test this to confirm, but I'm pretty sure if you're trying to validate your application will build correctly in stage 4 (for example), and you don't need to rebuild earlier stages, you can add SKIP files to stage 1, 2, and 3.

README.md
```diff
@@ -80,6 +80,48 @@ IMG_NAME='Raspbian'
...
+       to configure things like locale, etc.
+
+     - **00-packages** - A list of packages to install. Can have more than one, space
+       separated, per line.
+
+     - **00-packages-nr** - As 00-packages, except these will be installed using
+       the ```--no-install-recommends -y``` parameters to apt-get
+
+     - **00-patches** - A directory containing patch files to be applied
+
```
jcwren: Do you know what order the scripts are processed in? If there's both a 00-run.sh and 00-run-chroot.sh, which gets executed first? This question applies to the all the directories. Knowing the order of evaluation may be useful to some.

README.md
bjpirt: That's what it says a couple of lines down, but you need to build the earlier stages first to get the build directories in place

README.md
bjpirt: The 00 is just an example, they should really all be numbered uniquely

jcwren: I guess it would depend on if it's doing a sort of all entries, or looping through each type and sorting those. If it's the former, then indicate that no number should be reused in any name in that directory. If it's processing xx-patches before xx-run-chroot.sh, indicate that patches are done first, then xx-run.sh, then xx-run-chroot.sh, etc.

Your additions are good stuff, so please don't think I'm bashing on them (although we do use bash to run them :))

PR #137, opened 2017-12-18
Allow image building to be skipped for stages

bjpirt:
Whilst working on a fork of this repo to build my own images I added this modification which allows you to skip exporting images for stages (in the same way you can just skip a stage with SKIP files). There are two main reasons for wanting to do this;
 - it means you can make fewer changes to a fork of the repo and keep it in sync more easily and don't accidentally commit the removal of the export file
 - the EXPORT_* files aren't just empty but contain configuration so this makes it easier to just turn on and off image building during development

It's a pretty straightforward change to the build script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ work/*
 config
 postrun.sh
 SKIP
+SKIP_IMAGES
 .pc
 *-pc
 apt-cacher-ng/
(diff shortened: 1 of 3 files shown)


XECDesign:
Looks okay. I'll leave it open for a bit in case somebody has suggestions.

XECDesign:
Thanks

PR #139, opened 2017-12-26
Enable UART by default

Alkass:


XECDesign:
UART is enabled by default on models which don't require a fixed core clock. Forcing UART on all models can negatively impact performance.

PR #154, opened 2018-01-25
make root filesystem 4M aligned

jamesruan:
For most SD card has 4M erase block. It is good to align file system at 4M boundary.


XECDesign:
Thanks

PR #163, opened 2018-03-13
Update stage specification using SKIP_IMAGES

junian:
Change Stage specification guide from removing `EXPORT*` files to adding `SKIP_IMAGES` as per new recommended development process.


XECDesign:
Thanks

PR #168, opened 2018-03-24
Remove quotes in globbed patch CLEANup

davesteele:
Globbing does not work within single or double quotes.

Because of this, at least some patches were not being applied on rebuilds, regardless of the CLEAN setting.

Note that the shellcheck commit that introduced the quotes (ff2d5ed) also quoted [three globs](https://github.com/davesteele/pi-gen/blob/comitup/export-image/04-finalise/01-run.sh#L26) in export.


XECDesign:
Thanks

PR #170, opened 2018-03-24
Add README detail on the 'patches' process

davesteele:
Managing patches is much easier once quilt and EDIT are introduced.


XECDesign:
Thanks again

PR #172, opened 2018-03-25
Fix qemu mode

jacen92:
This reverts commit 56da271499afe7ae8e35a8fd30f04135283b2fb0.

Since this commit the qemu image crashes. I've revert it to fix but it is not the smartest way to fix the problem just the easier.


XECDesign:
That's on a standard build of stretch?

jacen92:
yes on master. When I tried to test the new image (for the pi3+) on Qemu (QEMU_MODE=1) it crashed.
I don't know exactly why but when I reversed this commit it works again.

PR #173, opened 2018-03-26
Fix some quoted globs in export-image cleanup

davesteele:
A few more bash globs were being disabled by quotes.


XECDesign:
Thanks again.

PR #177, opened 2018-04-03
Update README.md

voipminic:
apt-transport-https is required too

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ earlier releases of these systems.
 To install the required dependencies for pi-gen you should run:
 
 ```bash
-apt-get install quilt parted realpath qemu-user-static debootstrap zerofree pxz zip \
-dosfstools bsdtar libcap2-bin grep rsync xz-utils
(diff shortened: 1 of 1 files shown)


XECDesign:
Why would that be required?

r4and0m0:
I had to install that package too on latest xubuntu lts some time ago, the script just complained about the missing package and aborted the job.
BTW
At this moment the repo still have the old version of rpigen ? it generate just 2 img, lite and full(whitout mathematica)

XECDesign:
sorry @r4and0m0, didn't see your message before.

I think the repo should be up to date now and have all the variants.

I can't find anything in pi-gen that requires  apt-trasport-https. I'm open to being proven wrong, but with a full explanation of what's going on, because I'm not seeing it.

r4and0m0:
For me the script complained about the missing package apt-transport-https then stopped (happen like all other dependecys(is missing)) so when i installed the https apt transport package i was able to build any image.

PR #178, opened 2018-04-06
Custom config

jacen92:
This PR allows developpers to customize their image.
They have to set their values to the config file.

`cp config.exemple config`


Landrash:
Hoping this will be adopted since it will minimize the customization to the build script for the Hassbian image even more.

xoronet:
Great! I would also prefer such a solution

jacen92:
By default it will create images for qemu if you want to test it quickly.

jacen92:
Hi,

I rebased the branch could you please merge it?

XECDesign:
I feel like this is coupling pi-gen too closely to Raspbian. It should be fairly agnostic to what debian-based distro it's building or how it's set up. Also, where would this approach stop? I can think of a dozen other things that could arguably be set through .config.

On one hand, this doesn't fit with the idea behind pi-gen. On the other hand, if people don't have to change 20 files for minor changes, it's more user friendly. I don't know, I'm not sold on this yet.

Aside from that, why Revert "Remove unnecessary files"? There are also a few typos here and there, like go instead of gb and exemple instead of example.

jacen92:
You are right Revert "Remove unnecessary files" was a mistake and I removed it from this branch.
Personnaly I uses this features to create  custom images following my needs.
For example I have a config file to create a "test" image (with an other set of password, other hostname) and a "production" image (with a secret config file).
When my raspberry boot I know on which image I am with the custom hostname.
I admit it will open the door to more changes but I think it could be a good idea to have the same repository to build an image but with our personnal config file without the need to fork it.

Thank you
I invite you to take a look to my repository to see how I expanded this feature.

XECDesign:
pi-gen isn't meant for customising raspbian images, but for building debian-based distro images in general. I hate rejecting contributions, especially good ones like this, but at the end of the day, I have to use pi-gen the most and maintain it. I have to go with simplicity and let the purpose of pi-gen drive the development of it, rather than add features which would be nice, but add complexity and aren't going to be used for the primary purpose of the tool.

PR #186, opened 2018-05-09
export-image: clean additional backup files

hhromic:
Hi again, here is another small improvement.
I noticed that in recent builds there are two backup files (`/etc/sub{u,g}id-`) that aren't cleaned in the export-image -> finalize substage.
This PR fixes adds those files to the cleaning phase.


XECDesign:
Thanks

PR #187, opened 2018-05-09
stage2: clean unused/obsolete patch

hhromic:
Yet another little clean-up to pi-gen.
This time, I discovered that the `06-rc_local.diff` patch in stage2 -> sys-tweaks is dangling as can be seen in the corresponding `series` file.
This PR removes the obsolete/unused patch file.


XECDesign:
Good catch

Issue #183, opened 2018-04-29
Docker Stage 0 fail: debootstrap directory not empty.


moriczgergo:
I'm trying to build my own custom image.

It seems that a debootstrap.log remains in the debootstrap dir, and rmdir can't remove the directory because of that.

If I remove it, it fails on /etc/apt not existing.

```
update-binfmts: warning: qemu-aarch64 already enabled in kernel.
[10:10:07] Begin /pi-gen
[10:10:07] Begin /pi-gen/stage0
[10:10:07] Begin /pi-gen/stage0/prerun.sh
I: Running command: debootstrap --arch armhf --foreign --components=main,contrib,non-free --keyring /pi-gen/stage0/files/raspberrypi.gpg stretch /pi-gen/work/2018-04-29-Fusee Framboise/stage0/rootfs http://raspbian.raspberrypi.org/raspbian/
I: Retrieving InRelease
I: Checking Release signature
I: Valid Release signature (key id A0DA38D0D76E8B5D638872819165938D90FDDD2E)
rmdir: failed to remove '/pi-gen/work/2018-04-29-Fusee Framboise/stage0/rootfs/debootstrap': Directory not empty
```

hongquan:
I don't have space in IMG_NAME, but still encounter this "debootstrap directory not empty" error.

The `rmdir` command was used for deleting the directory, but my directory has 3 files inside, so `rmdir` refused to do its job.

OneStromberg:
@hongquan so how the issue was solved?

hongquan:
@OneStromberg I no longer use Docker build. It seems to be that the build script used wrong version of debootstrap. I already go back to traditional method (not through Docker).

PR #190, opened 2018-06-05
Update Readme to warn about IMG_NAME spaces

danShumway:
May prevent issues like #183 in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,10 @@ The following environment variables are supported:
 
  * `IMG_NAME` **required** (Default: unset)
 
-   The name of the image to build with the current stage directories.  Setting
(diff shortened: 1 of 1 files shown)


danShumway:
I got bitten by a similar issue, I assumed that since Raspbian was capitalized spaces would be OK.

Easy enough to find the answer, but adding a line in the documentation may save someone a few minutes in the future.

XECDesign:
It would be nicer to get to the root of the issue and check why IMG_NAME with spaces is a problem.

danShumway:
It's kind of a pain that there's no way in bash to permanently escape spaces in one place so that it's not reliant on other commands to use quotation marks properly. Maybe there is a way to do that and I've just never found it.

Will poke around :)

PR #196, opened 2018-07-06
Adding file to packages

KalleDK:
Build a clean image, but build.sh complains that 'file' is missing from the system.


XECDesign:
Thanks

Issue #133, opened 2017-11-13
failed to run command '/debootstrap/debootstrap': Exec format error


attie:
Trying to run on an x86-64 machine, running Ubuntu 16.04 with Docker 1.12.6, on commit 31fcc80.

My config:
```
$ cat config
IMG_NAME='my_project'
USE_QEMU=1
```

Interesting output is below:
```
$ ./build-docker.sh
Sending build context to Docker daemon 837.1 kB
Step 1 : FROM debian:stretch
 ---> 6d83de432e98
[...]
I: Extracting xz-utils...
I: Extracting zlib1g...
I: Running command: chroot /pi-gen/work/2017-11-13-my_project/stage0/rootfs /debootstrap/debootstrap --second-stage
chroot: failed to run command '/debootstrap/debootstrap': Exec format error
rmdir: failed to remove '/pi-gen/work/2017-11-13-my_project/stage0/rootfs/debootstrap': Directory not empty
```

I have confirmed that `qemu-debootstrap` is selected (according to `scripts/common:13`).

It looks like something is trying to run `/debootstrap/debootstrap` (a shell script) outside of QEMU - the script's shebang calls on `/bin/sh` which is a symlink to `/bin/dash` which is in turn an ARM binary (as below, from inside a container):
```
$ cd /pi-gen/work/2017-11-13-my_project/stage0/rootfs
$ head -n 1 ./debootstrap/debootstrap
#!/bin/sh
$ ls -l ./bin/sh
lrwxrwxrwx 1 root root 4 Jan 24  2017 ./bin/sh -> dash
$ file ./bin/dash
./bin/dash: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=18bddc2d67c785bbaaf97cbb2c0dd5420a1861b3, stripped
```

Can anyone help get this running?
Am I missing something?

Thanks!

attie:
Full output is below:
```
$ ./build-docker.sh
Sending build context to Docker daemon 837.1 kB
Step 1 : FROM debian:stretch
 ---> 6d83de432e98
Step 2 : ENV DEBIAN_FRONTEND noninteractive
 ---> Using cache
 ---> b212f8354864
[...]

PR #203, opened 2018-08-03
Added Troubleshooting section to README.md, and addressed binfmt_misc…

attie:
People have requested that potential binfmt_misc issues are raised in README.md (see #133).

This patch addresses this.


XECDesign:
Thanks

PR #214, opened 2018-10-20
Add curl to apt-get script in README

kr15h:
Noticed that the apt-get script does not install everything. Added curl to the apt-get install list.


XECDesign:
Thank you

PR #215, opened 2018-10-26
Only track non-skipped stages in `PREV_` variables

mmalecki:
Only set `PREV_STAGE`, `PREV_STAGE_DIR` and `PREV_ROOTFS_DIR` when the
stage is actually built.

When building a custom image, I wanted to make the least modifications
to this repo, so I've added `SKIP` files to `stage3` and `stage4`, and
created my own `stage5`, but this confused `copy_previous`, which was
trying to copy the rootfs from a un-built stage. This fixes this issue.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -108,11 +108,11 @@ run_stage(){
 				run_sub_stage
 			fi
 		done
+		PREV_ROOTFS_DIR="${ROOTFS_DIR}"
+		PREV_STAGE="${STAGE}"
+		PREV_STAGE_DIR="${STAGE_DIR}"
 	fi
 	unmount "${WORK_DIR}/${STAGE}"
-	PREV_STAGE="${STAGE}"
-	PREV_STAGE_DIR="${STAGE_DIR}"
-	PREV_ROOTFS_DIR="${ROOTFS_DIR}"
 	popd > /dev/null
 	log "End ${STAGE_DIR}"
 }


XECDesign:
Would this break the ability to skip all stages and just export them?

rtgoodwin:
I was playing about with a similar solution due to similar thought pattern: "keep my repo as close to master as possible for future merges".

The Readme suggests (or I mistakenly read it this way, which is quite certainly a possibility) that SKIP means "Skip building this no matter what". Instead it currently means "Skip building this _again_" with the assumption that it already built once.  (I know this probably seems obvious but just putting my newb impression out there.)

A _very_ quick test of this patch seems to work. Ignore the echo of `ROOTFS_DIR`...that was from me learning. 😄 
image-23-11-28
image-23-12-26

matthijskooijman:
> The Readme suggests (or I mistakenly read it this way, which is quite certainly a possibility) that SKIP means "Skip building this no matter what". Instead it currently means "Skip building this again" with the assumption that it already built once. (I know this probably seems obvious but just putting my impression out there.) I feel like the latter behavior is what might be generally expected?

I'm reading the README that SKIP just means "do not build this now", and can be used both for "never build this" (if it was not built before) and "Do not build again" (if it was built already). E.g. there are examples for only building the lite image by SKIPping stage3 and stage4 and there is an example that says to add SKIP to already built stages to speed up development.

This PR would, AFAICS, break the latter usecase of skipping already built stages, hence I do not think this PR is really the right way (unless the SKIP mechanism is extended to properly support both usecases).

However, there is, AFAICS, already a way to achieve with @mmalecki wants. By adding a `STAGE_LIST` to `config`, only the listed stages are considered at all and any directories not in the list are completely ignored.

mmalecki:
@matthijskooijman you're 100% right about `STAGE_LIST` being the preferred solution here. (Recent changes to `pi-gen` almost completely eliminated my need for a custom stage, too.)
I'll close this PR.

(Wow, almost 2 years since I had to build this image last, judging by the PR age.)

PR #219, opened 2018-11-08
scripts: do not ignore errors inside on_chroot calls

### LINKED ISSUE #217: Failures inside on_chroot do not stop the build
Most build scripts have "-e" option which is meant to make failed commands interrupt the script.

However, if any commands fail inside `on_chroot` call, the build still continues.

This seems a bit risky as you can only notice failures by carefully analysing build logs.

Replacing all instances of `on_chroot` with `on_chroot -e` fixes this.

tomas-c:
Fixes #217

### DIFF
--- a/scripts/common
+++ b/scripts/common
@@ -88,7 +88,7 @@ on_chroot() {
 		mount --bind /sys "${ROOTFS_DIR}/sys"
 	fi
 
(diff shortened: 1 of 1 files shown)


XECDesign:
Thank you

PR #224, opened 2018-12-09
Allow setting user name and password from config

samtygier:
Add FIRST_USER_NAME and FIRST_USER_PASS variables that can be set in the config (or enviroment). Defaults to the standard pi and raspberry.

### DIFF
--- a/stage2/01-sys-tweaks/01-run.sh
+++ b/stage2/01-sys-tweaks/01-run.sh
@@ -32,12 +32,12 @@ systemctl enable resize2fs_once
 EOF
 fi
 
-on_chroot << \EOF
+on_chroot <<EOF
 for GRP in input spi i2c gpio; do
-	groupadd -f -r "$GRP"
+	groupadd -f -r "\$GRP"
 done
 for GRP in adm dialout cdrom audio users sudo video games plugdev input gpio spi i2c netdev; do
-  adduser pi $GRP
+  adduser $FIRST_USER_NAME \$GRP
 done
 EOF
 
--- a/stage2/01-sys-tweaks/01-run.sh
+++ b/stage2/01-sys-tweaks/01-run.sh
@@ -15,7 +15,11 @@ on_chroot << EOF
 systemctl disable hwclock.sh
 systemctl disable nfs-common
 systemctl disable rpcbind
-systemctl disable ssh
+if [ "${ENABLE_SSH}" == "1" ]; then
+	systemctl enable ssh
+else
+	systemctl disable ssh
+fi
 systemctl enable regenerate_ssh_host_keys
 EOF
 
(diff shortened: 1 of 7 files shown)


samtygier:
This simplifies the common task of creating an image with a modified initial user account.

XECDesign:
Given that this is a popular request, alright.

Would it make more sense to export the variable in build.sh so that you don't have to have the same line (`USERNAME=${FIRST_USER_NAME:-pi}`) every time we need to use the variable?

samtygier:
Yes, that would be neater. Do you have any preference for the variable names?

Also would you be open adding to more customisation options? For example initial wifi setup or installing additional packages? Or do you think that would be better as an external project.

XECDesign:
No preference.

Again, the idea that the image should be more configurable without having to modify the stages themselves is one that comes up from time to time and I think I've been worn down to being open to it.

Originally, there was meant to be some separation between the stages to build raspbian and pi-gen itself. pi-gen being a generic tool to build debian-based images which may or may not have new users. But given that the main use seems to be to customise raspbian, it makes sense to add more options.

jcwren:
Perhaps a single file with all the configuration options to be supported, that's sourced by the build script, with the name of the file being able to be specified on the command line?
```
IMAGE_NAME=MySpecialImage.img
LOCALE=US
SSH_ENABLE=yes
ADD_PI_USER=no
DEFAULT_USER=newbie
DEFAULT_PASSWORD=not_a_good_password
<etc>
```

XECDesign:
That would work. So default to using the existing `config` file and allow passing different configs? No problem with that.

samtygier:
Here is the revised username/password options. I shall prepare patches for wifi keys and enabling ssh. Shall I add the extra config file option with an option like:

`./build.sh -c myconfig`

stage2/01-sys-tweaks/01-run.sh
```diff
@@ -32,12 +32,12 @@ systemctl enable resize2fs_once
...
 
-on_chroot << \EOF
+on_chroot <<EOF
 for GRP in input spi i2c gpio; do
-	groupadd -f -r "$GRP"
+	groupadd -f -r "\$GRP"
 done
 for GRP in adm dialout cdrom audio users sudo video games plugdev input gpio spi i2c netdev; do
-  adduser pi $GRP
+  adduser $FIRST_USER_NAME \$GRP
```
jcwren: Shouldn't `$FIRST_USER_NAME` (and any other of the strings) be double-quoted in case the string contains special characters?

samtygier: For the username it should be limited to a small set of characters anyway. So its best to just check that at the beginning than get so far an pass an invalid username to adduser. See  #235

samtygier: The wpa essid and password needed quoting as well. I added a second patch to #235

maksim-ka:
maybe the file /etc/sudoers.d/010_pi-nopasswd could be involved as well.

PR #235, opened 2019-01-05
Add username validation

samtygier:
Check that the username is valid before doing any work. Use the default regex from debian's adduser.conf. Will also avoid risk of special characters causing issues.


XECDesign:
Thank you

PR #236, opened 2019-01-06
Allow custom stages

samtygier:
STAGE_LIST can be specified in the config file to run additional build stages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,10 @@ The following environment variables are supported:
 
    Setting to `1` will enable ssh server for remote log in. Note that if you are using a common password such as the defaults there is a high risk of attackers taking over you RaspberryPi.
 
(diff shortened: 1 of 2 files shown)


samtygier:
I know you don't want to add every possible customization option that anyone can think of, but maybe this is a simple way to allow modular extensions. If it is a step to far, I could maintain a forked version of pi-gen. Thanks.

XECDesign:
I've mostly given up on that. The idea was that I'd work on pi-gen 2 which would take into account all of the feature requests and problems people have had, but I don't know when or whether I'll get around to it. So if people want to make pi-gen more useful for them and it doesn't cause any problems for Raspbian builds, I'm all for it.

Thanks for the contribution.

PR #237, opened 2019-01-14
Fixed a hard reference to the 'pi' user name

rkubes:
There was a recent commit to add the "FIRST_USER_NAME" option to config. In the export-image scripts, there was a section that still had a hard reference to the "pi" user name and did not respect the new "FIRST_USER_NAME" variable.

This pull request addresses that issue.


XECDesign:
Thanks

PR #240, opened 2019-01-18
Fixed realpath dependancy

Hereath:
On Debian 9, the package realpath is listed as *transitional package* ([see here](https://packages.debian.org/stretch/realpath)), and coreutils is the correct package to install it. Also on Ubuntu 18.04, realpath is not a standalone package in the repo, and is part of coreutils.

I've just updated the `README` and the `depends` file accordingly. The `dependancies_check()` will report the correct package to be installed (coreutils).

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ earlier releases of these systems.
 To install the required dependencies for pi-gen you should run:
 
 ```bash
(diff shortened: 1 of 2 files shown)


XECDesign:
Thank you

PR #243, opened 2019-01-21
IMG_SUFFIX evaluation in export image (fix #241)

### LINKED ISSUE #241: Images overwritten. IMG_SUFFIX not used
When I build multiple images at a time, like lite, dektop, qemu, all build go OK, images get generated but finally I get one image. It has no suffix like "lite", looks like every next image is consequently overwriting the previous one. 

build.sh is exporting something like this:
`export IMG_FILENAME="${IMG_FILENAME:-"${IMG_DATE}-${IMG_NAME}${IMG_SUFFIX}"}"`

but at this point IMG_SUFFIX is not yet defined. it  is going to be defined at EXPORT_IMAGE file, but it is never used later.

so file ends up without suffix and the last built image wins overwriting all previous ones

bjsowa:


XECDesign:
Thanks again

PR #250, opened 2019-01-30
Add a DEPLOY_ZIP setting

LouisMT:
Currently, the produced `.img` file is zipped and copied to the deploy directory.

Of course this saves a lot of space when transferring the image, so I understand that it is an useful feature. However:
* When I'm building images locally I don't need the zipped image and the zipping/unzipping consumes a lot of time.
* Automatically building images using (GitLab) CI already compresses the artifacts, resulting in nested `.zip` files.

This PR adds an option to disable zipping the final image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,10 @@ The following environment variables are supported:
 
    Output directory for target system images and NOOBS bundles.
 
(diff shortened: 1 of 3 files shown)


XECDesign:
Looks good. Thanks.

PR #252, opened 2019-01-30
Decrease zerofree verbosity

### LINKED ISSUE #251: Verbosity of zerofree
When generating images using a non-interactive shell, the [`zerofree -v` command](https://github.com/RPi-Distro/pi-gen/blob/61a994c88ceb4a9ded0aca5eeedbe56c801b5acd/export-image/04-finalise/01-run.sh#L71) generates hundreds of lines like this:

```
0.01%
0.02%
0.03%
0.04%
...
```

These lines are not very informative and burry useful information in the build log.

I would create a PR to remove the `-v` flag, but I assume there's a reason this flag was added in the first place? It would be nice if it could be removed, if the information is not really useful anyway. Otherwise maybe I can add a `-v` flag to `build.sh`, so verbosity can be controlled per build?

LouisMT:
Fixes #251.


XECDesign:
Thanks

PR #256, opened 2019-02-15
Improve flexibility of two script common functions

hhromic:
Improve bootstrap() function in common script:

* Tidy up and simplify function code
* Allow to pass extra arguments to debootstrap, e.g. `--variant minbase`
* Preserve compatibility with existing function calls

Improve update_issue() function in common script:

* Allow to customise pi-gen script name and repository in the generated issue file
* Preserve compatibility with existing function calls

Tested a complete distro build with these changes as well, all building fine here.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -137,6 +137,9 @@ do
 	esac
 done
 
+export PI_GEN=${PI_GEN:-pi-gen}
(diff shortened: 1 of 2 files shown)


XECDesign:
Thanks, much appreciated.

Issue #193, opened 2018-06-12
Build not successful as the file is not found


Chan9390:
**Tested on:** Debian 9

I have tried to build the image multiple times using `build.sh` but it fails at the end. I have followed the instructions as per those mentioned in the readme file.  The installation and compilation part is successful but when it comes to exporting the image, the build fails. 

The build log is as follows:

```
[13:19:11] Begin /builds/user/pi-gen/export-image/prerun.sh
Welcome to fdisk (util-linux 2.29.2).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

Device does not contain a recognized partition table.
Created a new DOS disklabel with disk identifier 0xe86e87dd.

Command (m for help): Created a new DOS disklabel with disk identifier 0x76b65104.

Command (m for help): Partition type
   p   primary (0 primary, 0 extended, 4 free)
   e   extended (container for logical partitions)
Select (default p): 
Using default response p.
Partition number (1-4, default 1): First sector (2048-3514367, default 2048): Last sector, +sectors or +size{K,M,G,T,P} (8192-3514367, default 3514367): 
Created a new partition 1 of type 'Linux' and of size 43.1 MiB.

Command (m for help): Disk /builds/user/pi-gen/work/2018-06-08-exos/export-image/2018-06-08-exos-lite.img: 1.7 GiB, 1799356416 bytes, 3514368 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x76b65104

Device                                                                              Boot Start   End Sectors  Size Id Type
/builds/user/pi-gen/work/2018-06-08-exos/export-image/2018-06-08-exos-lite.img1       8192 96484   88293 43.1M 83 Linux

Command (m for help): Selected partition 1
Partition type (type L to list all types): Changed type of partition 'Linux' to 'W95 FAT32 (LBA)'.

Command (m for help): Partition type
   p   primary (1 primary, 0 extended, 3 free)
   e   extended (container for logical partitions)
Select (default p): 
Using default response p.
Partition number (2-4, default 2): First sector (2048-3514367, default 2048): Last sector, +sectors or +size{K,M,G,T,P} (98304-3514367, default 3514367): 
Created a new partition 2 of type 'Linux' and of size 1.6 GiB.

Command (m for help): 
Command (m for help): 
[...]

Issue #104, opened 2017-09-01
running out of loopback devices, then rsync issues.


timcoote:
I'm running pi-gen's build-docker.sh in various fedora vms. I find that I need to create extra loop devices on the guest vm, otherwise docker containers seem to run out in `export-image/prerun.sh`, near here:
`BOOT_DEV=$(losetup --show -f -o ${BOOT_OFFSET} --sizelimit ${BOOT_LENGTH} ${IMG_FILE})
ROOT_DEV=$(losetup --show -f -o ${ROOT_OFFSET} --sizelimit ${ROOT_LENGTH} ${IMG_FILE})
`
like this:
```
+ ROOT_LENGTH=2988912640
++ losetup --show -f -o 4194304 --sizelimit 43854848 /pi-gen/work/2017-09-01-osx/export-image/2017-09-01-osx-4GB.img
losetup: /pi-gen/work/2017-09-01-osx/export-image/2017-09-01-osx-4GB.img: failed to set up loop device: No such file or directory
+ BOOT_DEV=
```
Clearly there's a workaround, eg:
```
# create a couple of dummy files:
dd if=/dev/zero of=virtualfs1 bs=1024 count=30720
dd if=/dev/zero of=virtualfs1 bs=1024 count=30720
# incrementally associate the files with loop devices
sudo losetup -f    # creates loop0
sudo losetup /dev/loop0 virtualfs1
sudo losetup -f   # creates loop1
sudo losetup /dev/loop0 virtualfs2
# remove the associations:
sudo losetup -d /dev/loop0
sudo losetup -d /dev/loop1
```
I don't know whether this is a specific redhat distro issue. I thought it worth pointing out. It would be a pita to automate, I think.

Once the script gets onto the rsync command (`rsync -aHAXx --exclude var/cache/apt/archives ${EXPORT_ROOTFS_DIR}/ ${ROOTFS_DIR}/`), there seems to be some tangling with SELinux as the -X flag (--xattrs) fails for the vfat partitions (/boot and /dev), with errors like this:
```
output.build.osx.7:+ rsync -aHAXx --exclude var/cache/apt/archives /pi-gen/work/2017-09-01-osx/stage4/rootfs/ /pi-gen/work/2017-09-01-osx/export-image/rootfs/
output.build.osx.7:rsync: rsync_xal_set: lsetxattr(""/pi-gen/work/2017-09-01-osx/export-image/rootfs/boot"","security.selinux") failed: Operation not supported (95)
[...]

PR #258, opened 2019-02-16
Make sure unmount_image finds stale loop devices

### LINKED ISSUE #257: Docker: losetup failed to set up loop device: No such file or directory
When building with docker I am getting the error:

`losetup: /pi-gen/work/2019-02-13-Raspbian/export-image/2019-02-13-Raspbian-lite.img: failed to set up loop device: No such file or directory`

even though that file does exist.

To reproduce, make docker build stop at final rsync in export-image/prerun.sh . Loop devices are at kernel level outside the docker container, so they persist after the failure. They can be seen with the command `losetup --list`

The unmount_image function in scripts/common uses:
LOOP_DEVICES=$(losetup -j "${1}" | cut -f1 -d':')
but does not see the old loop devices, because the path is missing the "/pi-gen/work", so it is unable to remove them. I imagine the path difference is to do with the how the container views the filesystem. This can be solved by listing all the loop devices, and matching by filename (patch to follow).

There are a few older closed bugs mentioning the issue, #104 #193

samtygier:
Fixes #257 #104 #193

Instead of searching by full path, which is prone to fail, read full list and grep on filename.

### DIFF
--- a/scripts/common
+++ b/scripts/common
@@ -56,7 +56,7 @@ unmount_image(){
 	sync
 	sleep 1
 	local LOOP_DEVICES
-	LOOP_DEVICES=$(losetup -j "${1}" | cut -f1 -d':')
+	LOOP_DEVICES=$(losetup --list | grep "$(basename "${1}")" | cut -f1 -d' ')
 	for LOOP_DEV in ${LOOP_DEVICES}; do
 		if [ -n "${LOOP_DEV}" ]; then
 			local MOUNTED_DIR


XECDesign:
Wouldn't there be an issue with this potentially unmounting things it shouldn't? If you have two builds running in parallel, for example, matching by basename alone could cause issues.

I'm happy to merge this either way, since it fixes a bigger issue than it introduces, but it would be nice to find a better way.

samtygier:
build-docker.sh already calls `docker ps` to check that it there is not a container with the same name running.

I think 2 builds in parallel would already be a problem, if they have the same date and image name, then the old method would still be trying to unmount them as they would have the same path. If they have different image names, then the new version would not match them.

XECDesign:
> build-docker.sh already calls `docker ps` to check that it there is not a container with the same name running.

I don't use docker, so that wouldn't help.

> I think 2 builds in parallel would already be a problem, if they have the same date and image name, then the old method would still be trying to unmount them as they would have the same path. If they have different image names, then the new version would not match them.

AFAIK, it currently uses the full path to check whether the image is mounted. Either way, the last time I ran parallel builds was never, so I'm not going to complain.

Thank you.

PR #259, opened 2019-02-16
Ensure extra config file is passed to docker

### LINKED ISSUE #253: custom config with build-docker.sh
When using script to build image in docker `build-docker.sh` the `-c` command line argument to specify different config file does not seem to work.

samtygier:
Fixes #253

Rather than use --env-file, which gets overwritten by the regular config file, pass the any arguments to build-docker.sh through to build.sh.


XECDesign:
Thanks

PR #261, opened 2019-02-17
Fedora fixes

### LINKED ISSUE #260: Rsync fails with xattrs on files in boot
in export-image/prerun.sh the command
`rsync -aHAXx --exclude var/cache/apt/archives "${EXPORT_ROOTFS_DIR}/" "${ROOTFS_DIR}/"`
with
`rsync: rsync_xal_set: lsetxattr(""/pi-gen/work/2019-02-14-Raspbian/export-image/rootfs/boot"","security.selinux") failed: Operation not supported (95)`
if any files have extended attributes set, as they are not support on fat32. (or if they have different owners/groups see #119 ).

This will always happen when using a host with selinux running, as files get security labels.

Recent version of rsync (3.1.3) allow filtering with `--filter='-X security.selinux'`, but that would require using debian buster in the docker.

Attributes are needed for some files in root, e.g. the ping binary.

A quick solution is to just use separate rsync calls for root and boot. (patch to follow)

samtygier:
While testing on fedora I made a few fixes (also tested on an Ubuntu host), on top of #258 #259

5f3ddb2 Check binfmt_misc module is loaded
Gives a friendlier message when binfmt_misc is not enabled

1c9df06 Use different rsync options for boot and root
Fixes #260

fc3d814 README.md: Add more info on Docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get -y update && \
     apt-get -y install \
         git vim parted \
         quilt realpath qemu-user-static debootstrap zerofree pxz zip dosfstools \
-        bsdtar libcap2-bin rsync grep udev xz-utils curl xxd file \
+        bsdtar libcap2-bin rsync grep udev xz-utils curl xxd file kmod\
     && rm -rf /var/lib/apt/lists/*
 
 COPY . /pi-gen/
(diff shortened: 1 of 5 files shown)


XECDesign:
Thanks

PR #279, opened 2019-04-24
Silence shellcheck warnings

hhromic:
* SC1091: the `config` file might not be present, which is normal.
* SC2086: Double quote to prevent globbing and word splitting.

Tested clean output using: `find -name "*.sh" -exec shellcheck -x {} \;`


XECDesign:
Thank you

PR #280, opened 2019-04-24
Do not assume the changelog file is always present for the info file

hhromic:
The `(..)/raspberrypi-kernel/changelog.Debian.gz` file is not guaranteed to be present in the built `ROOTFS_DIR`, for example when building very minimal images without package documentation. In these cases, the `firmware` variable will be left empty and the subsequent calls to `curl` will return large 404 HTML content from GitHub.

Instead, simply check if the changelog file exists before using it.

This does not affect the standard image generation for official Raspbian (tested).


XECDesign:
Thanks

PR #285, opened 2019-05-13
Use parted for partitioning

### LINKED ISSUE #284: Several issues with partitioning
I had a look at the image partitioning code and noticed some issues:
 1. There is some empty space between the boot and root partitions. Since commit 40eae05 (make root filesystem 4M aligned (#154)), the start of the root partition is aligned, but the boot partition size is unchanged, which leaves some unused space (which is a waste and makes it slightly harder to add a new partition in the remaining space on a disk, which is why I noticed this).
 2. There is an off-by-one in the aligning of the root partition. If the boot partition size is already exactly rounded to 4M increments, the root partition is put 4M after the boot partition, rather than adjacent to it. This is pretty much #156, but about a missing -1 in this line: https://github.com/RPi-Distro/pi-gen/blob/8acf95f237d3e0bbc5c698bcae2ad4f99efecccc/export-image/prerun.sh#L16
 3. Sending data into fdisk seems fragile to me. fdisk is, AFAIK, intended to be run interactively and the current script relies on default values and other interactive behaviour which could very well change in future versions.

I had a stab at fixing this with the current fdisk-based approach, which fixes 1) and 2), but obviously not 3). After doing so, I wondered if it would not be better to switch to parted for partition rather than fdisk.

One downside of parted is that, I think, it requires more specific instructions about the partition layout. It has no way to non-interactively say "put root adjacent to boot" without preculculating the exact starting point (note that specifying "use the rest of the disk" *is* possible by specifying `100%`). I suspect this is why fdisk was chosen originally. 

However, with the alignment for the root partition that is done now, pretty much the entire partition layout is calculated by `prerun.sh` already, so it seems sensible to use parted rather than fdisk (parted is already used to print the partition layout, so the dependency is already present).

If this sounds good, I can probably provide a PR to implement this somewhere in the coming days (after that, I'll probably be moving to other projects and might not have time anymore).

matthijskooijman:
Previously, fdisk was used by sending commands into its stdin, which is
not very robust (since it heavily relies on the interactive prompts
offered by fdisk as well as the default values it offers, which seem
prone to changing in future version).

It seems likely that in the past, fdisk was easier than parted since it
provides default values that make it easier to create adjacent
partitions, without precalculating all positions in the script. However
now that partitions are manually being aligned, all data must be
calculated anyway.

This commit changes the partition generation to use parted rather than
fdisk. For this, it rewrites various calculations and renames variables
to be easier to read as well. All values are now in number of bytes,
rather than mixing bytes and sectors.

This commit also makes makes sure that the boot partition and root
partition are always adjacent (previously the root partition was aligned
without also rounding the boot partition size, leaving some empty space
in between).

As a side effect of using parted, this also causes the "bootcode" part
of the MBR to be filled with some default x86 bootcode. This is totally
irrelevant for booting the Raspberry Pi, but it does prevent triggering
a bug in parted. When using parted to change the partition table (e.g.
when resizing the root partition on first boot by raspi-config's
init_resize.sh), the disk identifier would be changed due to this bug,
which would change the PARTUUID of all partitions. The init_resize.sh
script would work around this by updating the PARTUUID in e.g. fstab,
but that's fragile at best.  This commit prevents the bug from
triggering and keeps the disk identifier the same.

See https://debbugs.gnu.org/35714 for details about this parted bug.

This commit fixes #284.

### DIFF
--- a/export-image/prerun.sh
+++ b/export-image/prerun.sh
@@ -10,32 +10,27 @@ rm -rf "${ROOTFS_DIR}"
 mkdir -p "${ROOTFS_DIR}"
 
 BOOT_SIZE="$((256 * 1024 * 1024))"
-TOTAL_SIZE=$(du --apparent-size -s "${EXPORT_ROOTFS_DIR}" --exclude var/cache/apt/archives --block-size=1 | cut -f 1)
+ROOT_SIZE=$(du --apparent-size -s "${EXPORT_ROOTFS_DIR}" --exclude var/cache/apt/archives --exclude boot --block-size=1 | cut -f 1)
 
-ROUND_SIZE="$((4 * 1024 * 1024))"
-ROUNDED_ROOT_SECTOR=$(((BOOT_SIZE + ROUND_SIZE) / ROUND_SIZE * ROUND_SIZE / 512 + 8192))
-IMG_SIZE=$(((BOOT_SIZE + TOTAL_SIZE + (800 * 1024 * 1024) + ROUND_SIZE - 1) / ROUND_SIZE * ROUND_SIZE))
+# All partition sizes and starts will be aligned to this size
+ALIGN="$((4 * 1024 * 1024))"
+# Add this much space to the calculated file size. This allows for
+# some overhead (since actual space usage is usually rounded up to the
+# filesystem block size) and gives some free space on the resulting
+# image.
+ROOT_MARGIN=$((800*1024*1024))
+
+BOOT_PART_START=$((ALIGN))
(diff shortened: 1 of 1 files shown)


XECDesign:
Thanks for that. It may be a while before it gets merged, but at a glance, it looks much batter.

matthijskooijman:
Thanks! Take your time, I'm not blocked by this merge (I'm using a fork of pi-gen customized for my own image anyway).

matthijskooijman:
I just rebased this PR on top of master, now it applies without conflicts again.

XECDesign:
Sorry, this dropped off the radar for a bit. Thank you.

matthijskooijman:
No problem, thanks for merging!

PR #300, opened 2019-06-29
Remove obsolete IPv6 turnoff

oskar456:
Turning off IPv6 by aliasing `net-pf-10` to `off` does not work anymore.
Also, turning off IPv6 on system level breaks apps depending on IPv6
loopback or IPv6 link-local addresssing and creates issues in both
dual-stack and IPv6-only environments.

Signed-off-by: Ondřej Caletka <ondrej@caletka.cz>


XECDesign:
Thank you

PR #302, opened 2019-07-01
build-docker.sh does not assume PWD=<repo root>

macisamuele:
The `build-docker.sh` script should not assume that is executed from the repository root especially as the assumption is only used to build the docker-container.

Allowing the script to "auto-determine" its directory allows users to create configurations on their desired directories and then run something like `bash ${RPi-Distro/pi-gen/build-docker.sh} -c config`.


XECDesign:
Thanks

PR #306, opened 2019-07-06
Ensure that the configuration file is an absolute path in Docker build

hhromic:
PR #302 broke the Docker build script as previously documented. See below.

The specific problem is in commit 2ddd7c1c, where the passed config file (using the `-c` option) is now mounted inside the container using the `--volume src:dest:opt` Docker option.

If you pass a relative-path config file (like in the *non-working* example you gave in the PR description) you get this error after building the Docker image and starting the `pi-gen` build:

    bash ${RPi-Distro/pi-gen/build-docker.sh} -c config
    (...)
    ./build.sh: line 137: source: /config: is a directory

The problem is that Docker requires absolute paths for mounting single files inside the container, otherwise it silently tries to mount a volume name instead as an empty directory. Therefore the Docker build no longer works with the following invocation forms (relative config-paths):

    ./build-docker.sh -c myconfig
    /path/to/build-docker.sh -c myconfig   # also doesn't work

And instead the Docker build must now be invoked using absolute config-paths:

    ./build-docker.sh -c /path/to/myconfig   # works
    ./build-docker.sh -c $PWD/myconfig   # alternative
    /path/to/build-docker.sh -c /path/to/myconfig   # another alternative

~This is not really fixable because is just the way Docker works when using `--volume`.~

~Instead, this PR improves the documentation overall and adds how `build-docker.sh` works currently.~

**Update:** Fixed now using `realpath` (see commit 4fd20bd) that preserves old behaviour.

Also I'm silencing a new shellcheck warning and slightly fixing a missing `&&` in a Docker pipeline.

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -30,6 +30,9 @@ do
 	esac
 done
 
+# Ensure that the configuration file is an absolute path
+CONFIG_FILE=$(realpath -s "$CONFIG_FILE")
+
 # Ensure that the confguration file is present
 if test -z "${CONFIG_FILE}"; then
 	echo "Configuration file need to be present in '${DIR}/config' or path passed as parameter"
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -73,7 +73,7 @@ if [ "${CONTAINER_EXISTS}" != "" ]; then
 		--volumes-from="${CONTAINER_NAME}" --name "${CONTAINER_NAME}_cont" \
 		pi-gen \
 		bash -e -o pipefail -c "dpkg-reconfigure qemu-user-static &&
-	cd /pi-gen; ./build.sh ${BUILD_OPTS} ;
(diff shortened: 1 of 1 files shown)


XECDesign:
A few merge conflicts that need to be resolved

hhromic:
I have a better idea now to solve the issue addressed by this PR: use `realpath` to resolve the config file if given as relative. This way it will preserve the expected behaviour. I will be updating the PR in coming days after testing. Please keep it open.
Thanks!

hhromic:
@XECDesign I rebased and changed now the PR to use `realpath` in the Docker build script. In this way, the old behaviour is preserved therefore no extra documentation is necessary. It just works.
I tested this and is working fine again for all cases including relative config file arguments.
Hope you are fine with it otherwise let me know of any changes desired.

XECDesign:
Thanks again, Hugo. Much appreciated.

PR #308, opened 2019-07-06
Make build_docker.sh portable again

hilli:
Last commit made the script break on macOS.

From `man sed` (On Linux):

```
       -E, -r, --regexp-extended

              use extended regular expressions in the script (for portability use POSIX -E).
```


XECDesign:
Thank you

PR #326, opened 2019-08-27
Fix export-image to not depend on fake-hwclock being installed

rkubes:
Currently the `export-image/04-finalise/01-run.sh` script depends on `fake-hwclock` being installed, since it calls `/etc/init.d/fake-hwclock stop` early on.
Since it's run under `bash -e` the build fails if `fake-hwclock` is not installed.
This is not an issue for the majority of users, but it prevents images from being built prior to `stage2`, and would impact any customization that removes `fake-hwclock` prior to `export-image`.

I could have done this by checking if the `/etc/init.d/fake-hwclock` file exists before executing it; however, that seemed a little extra rather than just adding the " || true" at the end of the command to force a 0 exit code. The risk with my method is if `fake-hwclock` is actually installed, but for some reason fails to stop, then the `export-image` script will continue anyway. My opinion is that risk is minimal, but I can change the pull request if desired.


XECDesign:
Looks good, but maybe an `if [ -x /usr/bin... ]; then` type check is better to avoid unnecessary error being shown in the output?

XECDesign:
Thank you, much appreciated.

PR #328, opened 2019-09-23
Update config.txt to replace lirc-rpi with gpio-ir

fpicalausa:
The example for enabling IR transmission in `/boot/config.txt` is still using the deprecated `lirc-rpi` overlay. 
The documentation in `/boot/overlays/README` indicates that this overlay has been deprecated in favor of `gpio-ir` / `gpio-ir-tx`.

This updates the actual config.txt to suggest `gpio-ir` instead of `lirc-rpi`.


XECDesign:
Thank you

PR #330, opened 2019-09-25
Customize root access

jacen92:
This request contains a way to add a public key to the root user to access it via ssh (usefull for first configuration in server mode).

With this you just need to add in the config file the key ROOT_PUBLIC_KEY with your public key as value.

Next step: disabling ssh authentication with password for root user to improve security.

### DIFF
--- a/stage2/01-sys-tweaks/01-run.sh
+++ b/stage2/01-sys-tweaks/01-run.sh
@@ -21,6 +21,10 @@ else
 	systemctl disable ssh
 fi
 systemctl enable regenerate_ssh_host_keys
+if [ -n "${ROOT_PUBLIC_KEY}" ]; then
+	mkdir -p /root/.ssh
+	echo "${ROOT_PUBLIC_KEY}" >> /root/.ssh/authorized_keys
+fi
 EOF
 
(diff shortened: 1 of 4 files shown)


stage2/01-sys-tweaks/01-run.sh
```diff
@@ -21,6 +21,10 @@ else
 	systemctl disable ssh
 fi
 systemctl enable regenerate_ssh_host_keys
+if [ -z "${ROOT_PUBLIC_KEY}" ]; then
```
XECDesign: shouldn't that be `-n`?

jacen92: You are right it could be better ;)

Fixed

sneak:
It would be cool if this were worked into the existing support for the `ssh` file in the boot, and then the key could just be added to the config.txt in the boot partition in the same way, then you don't need the root ssh key at image building time.

cdupont:
This would be nice. Any plan to merge? Thanks

cdupont:
Actually I think the use case is covered by PUBKEY_SSH_FIRST_USER and PUBKEY_ONLY_SSH now.

PR #331, opened 2019-09-26
Install cloud-init

kraeml:
With installing cloud-init you could initialize your raspberry with the file /boot/user-data.

With the [flash](https://github.com/hypriot/flash) tool from hypriot team you could setup this  during the flash process.

### DIFF
--- a/stage2/02-cloud-init/00-packages
+++ b/stage2/02-cloud-init/00-packages
@@ -1,0 +1,2 @@
+cloud-init
+ssh-import-id
--- a/stage2/02-cloud-init/01-run.sh
+++ b/stage2/02-cloud-init/01-run.sh
@@ -1,0 +1,20 @@
+#!/bin/bash -e
+
+install -m 644 files/user-data ${ROOTFS_DIR}/boot/
+install -m 644 files/meta-data ${ROOTFS_DIR}/boot/
+install -m 644 files/cloud/cloud.cfg ${ROOTFS_DIR}/etc/cloud/
+
+on_chroot << EOF
+echo 'Get cloud-init'
+
+echo 'Symlinking cloud-init nocloud-net seed files to /boot partition'
+mkdir -p /var/lib/cloud/seed/nocloud-net
+if [ ! -f /var/lib/cloud/seed/nocloud-net/user-data ]
+then
+  ln -s /boot/user-data /var/lib/cloud/seed/nocloud-net/user-data
+fi
+if [ ! -f /var/lib/cloud/seed/nocloud-net/meta-data ]
+then
+  ln -s /boot/meta-data /var/lib/cloud/seed/nocloud-net/meta-data
+fi
+EOF
--- a/stage2/02-cloud-init/files/cloud/cloud.cfg
+++ b/stage2/02-cloud-init/files/cloud/cloud.cfg
@@ -1,0 +1,187 @@
+# The top level settings are used as module
+# and system configuration.
+
+# A set of users which may be applied and/or used by various modules
+# when a 'default' entry is found it will reference the 'default_user'
+# from the distro configuration specified below
+users:
+   - default
+
+# If this is set, 'root' will not be able to ssh in and they
+# will get a message to login instead as the above $user (debian)
+disable_root: true
+
+# This will cause the set+update hostname module to not operate (if true)
+preserve_hostname: false
+
+# This preserve the sources list
+apt_preserve_sources_list: true
+
+datasource_list: [ NoCloud ]
+
+# Example datasource config
+#datasourcef:
+#    Ec2:
+#      metadata_urls: [ 'blah.com' ]
+#      timeout: 5 # (defaults to 50 seconds)
+#      max_wait: 10 # (defaults to 120 seconds)
+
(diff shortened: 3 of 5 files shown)


XECDesign:
For what it's doing, it seems to be way more complicated than it needs to be. We have an alternative approach in the pipeline for this.

kraeml:
You can setup your raspbian with editing or flashing only one yaml file `/boot/user-data`. So you can setup the hostname, create new user with ssh-key, install individual packages, setup wifi before running respbian and so on. Its very cool. 

Have a look at https://cloudinit.readthedocs.io/en/latest/

_Cloud-init is the industry standard multi-distribution method for cross-platform cloud instance initialization. It is supported across all major public cloud providers, provisioning systems for private cloud infrastructure, and bare-metal installations._

Bare-metal is her the Pi.

I'm teacher and when I setup 20 Pis I have 20 rapsberrypis hostname in my network. So I could flash it with an individual hostname for every student.

XECDesign:
I can definitely see the merit of it. It's just that last time we had discussions about this internally the consensus was that it's a bit too much for the target audience and a simple self-documented text file with just the most common options was settled upon.

kraeml:
I agree with you that cloud-init with the modules for aws, azure etc. is overwhelmed. But you don't have to use the whole cloud-init stack only the nocloud part. A simple text file (Yaml is a simple text file) user-data with some examples are enough to configure an pi-cluster out of the box (out of the flash progress).

I'm a teacher and my benefit using cloud-init is:

* several pis have unique hostname  in network (Without knowing ip or mac)
* setting up rdp for headless
* set up users with various ssh-keys
* set up ansible via cloud-init (needed packages)
* clone a repro with an ansible-playbook and run it
* set up cloud9, jupyter, nodered with various npm modules via ansible
* clone some nodeschool lessons 
*  ...

 I don't know yours simple self-documented text file. But I think you can use it parallel with cloud-init. Is there any documentation of this text file? What are the most common options?

jriguera:
Some time ago, in a personal project, I thought about using cloud-init for automatic Raspberry Pi configuration, using `/boot` partition as **datasource**. After a few tries I get to the conclusion that is not the best way to auto-configure as raspberry pi:

* Requires some tuning. The package is not directly usable after installing it. You can see it in these files: https://github.com/jriguera/raspbian-cloud/tree/master/stage8/90-cloud-init/files
* Cloud-init keeps a state after the first boot of the Raspberry. So if a user, after the first boot, changes some settings in the `/boot` partition, maybe they are not going to be applied (it depends on the settings).
* The configuration is quite complex for regular users:
  * When it boots, it runs different phases.
  * Some modules need to be disabled because they are not suitable for a Raspberry Pi.

I have decided to create a tool  [confinit](https://github.com/jriguera/confinit) which basically is a template engine capable of running commands. The idea is creating a folder in `/boot/config` with configuration templates for files (following the same fs hierarchy)  and a configuration file with all variables used in those templates (`parameters.yml`). The user only has to setup variables on the `parameters.yml` file and boot the raspberry pi. When it boots, it will render the templates  based on the rules defined in the main configuration file and optionally run commands (even the template files can be run after creating them). Ofc, the folder with templates and the main configuration file can be hidden for the end users, but I opt to put everything in `/boot` to give more flexibility.

You can see a small example here: https://github.com/jriguera/confinit/tree/master/example
and a complete production example here: https://github.com/jriguera/raspbian-cloud/tree/master/stage8/99-confinit/config

Maybe it can help in your use case.

dehnert:
I'm curious whether things have changed at all here. My impression is that these days cloud-init is the de facto standard for provisioning. Given that, I think there's lots of value to supporting it (possibly in addition to other approaches) -- there's lots of documentation, blog posts, and StackOverflow answers to help people set it up and people with a mixture of RPis and VMs (or RPis running multiple distros) can use similar toolchains to bootstrap their systems (my personal motivation). The existence of https://github.com/timebertt/pi-cloud-init implies there's more demand for this, and getting it in the upstream distro so one doesn't need to use third-party images would be helpful.

I haven't tried this pull request out or started looking at the practical aspects, but if there's willingness in principal to accept a PR along these lines I could.

XECDesign:
The plan is to add cloud-init support when we get a chance, but a PR won't help.

While cloud-init is the de dafcto standard for servers, last time I checked, it doesn't quite do what it claims when it comes to the desktop. I've raised some issues in the past and I think they have been addressed upstream, so I'll need to do another pass and check that it solves the problems we need it to solve.

For now, all effort right now is on getting bookworm ready. Cloud-init support is not too far down the todo list after that (hopefully).

dehnert:
Now that Bookworm is out, I'm curious about whether there's any news on cloud-init support / timing?

(FWIW, my RPi's are servers / used headless, so from my perspective, "works for server stuff, but not so much for desktop" would be fine. I realize this is not universal, and I don't know what the user breakdown looks like overall.)

XECDesign:
Most, if not all the concerns I had about cloud-init have been resolved upstream. The dust still hasn't quite settled from the bookworm release and there are lots of issues to address there. I'm still hoping to revisit cloud-init Soon™.

dehnert:
Any news?

PR #333, opened 2019-10-11
BerryLan Mods (Raspbian buster)

mzanetti:
**Not for merging. This is just to show a diff for the berrylan mods**

* replace dhcpcd5 with network-manager
* seed in the nymea repository and nymea-networkmanager
* patch nymea-networkmanager startup comand to use "Raspberry Pi" instead of nymea


mzanetti:
oops. wrong repo. sorry

PR #323, opened 2019-08-13
Support regional config settings

jbussdieker:
Addresses some of the concerns in issue #212. For me it's mostly about fixing the issue where `|` on my keyboard incorrectly maps to `~`.

PR #334, opened 2019-10-14
Document how to get values for the localization values

matthijskooijman:
In #323, some extra localization config values were added. This PR adds some extra documentation for some of these values, showing how to get the current values from a running system.


XECDesign:
Thanks

PR #337, opened 2019-10-26
hostname is now configurable

reneklootwijk:
I make the hostname configurable via the config file using the HOSTNAME variable. If the HOSTNAME has not been set the default is 'raspberrypi', otherwise the hostname will be set to the specified value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ The following environment variables are supported:
    be built and cached.  Note, `WORK_DIR` stores a complete copy of the target
    system for each build stage, amounting to tens of gigabytes in the case of
    Raspbian.
-   
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -82,7 +82,11 @@ The following environment variables are supported:
 
    Default system locale.
 
- * `KEYBOARD_KEYMAP` (Default: "gb" )
+* `HOSTNAME` (Default: "raspberrypi" )
+
+   Setting the hostname to the specified value.
+
+* `KEYBOARD_KEYMAP` (Default: "gb" )
```
XECDesign: These don't align with the rest of the file

XECDesign:
I'll have to change `HOSTNAME` to something else, because bash sets the variable automatically. Then builds ends up inheriting the build system's hostname.

I'll go with `TARGET_HOSTNAME` unless somebody has a better suggestion.

PR #341, opened 2019-11-11
Platform fixes for build-docker.sh

snoe925:
Test for /usr/bin/realpath.  OS X does not normally have realpath.  User will have to pass an absolute path.

Support "sudo docker" for the docker command.  Not every platform is configured for docker build as non-root user.
```
DOCKER="sudo docker" bash build-docker.sh
```


XECDesign:
I'll leave this one for docker users to comment on.

PR #347, opened 2019-11-18
Export base directory before loading config file

LCvanDinteren:
By exporting the base directory before loading config, it is possible to use the base dir within the config file.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -122,6 +122,7 @@ if [ "$(id -u)" != "0" ]; then
 	exit 1
 fi
 
+export BASE_DIR=${BASE_DIR:-"$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"}
 
 if [ -f config ]; then
 	# shellcheck disable=SC1091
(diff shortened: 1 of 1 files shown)


XECDesign:
I suspect shellcheck wouldn't be happy with that and want the assignment and export on separate lines.

LCvanDinteren:
> I suspect shellcheck wouldn't be happy with that and want the assignment and export on separate lines.

This same structure is used for many exports below, which is why I chose to keep in line with those lines.
Do you suggest something like this? :
```
if [ -z "${BASE_DIR}" ]; then
  BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
fi
export BASE_DIR
```

XECDesign:
I didn't know about shellcheck when I wrote the first version, so I wouldn't be surprised. Just trying to stick to better practices now.

In most case, an export and assignment on the same line is fine because there's no command substitution. In this case, because the assignment can fail, you'd want to catch that (which doesn't happen if you export on the same line).

I'm not sure the `[ -z "${BASE_DIR}" ]` would make sense. What would be the use case for setting a BASE_DIR that's outside of build.sh? Has that use case been tested to work with that change or would it also be broken in other ways?

I think if you move the existing BASE_DIR= and export BASE_DIR lines up so that's they're after the root user check, it should be fine.

LCvanDinteren:
Alright, always a fan of better practices:)

I can't think of any use for defining BASE_DIR outside of build.sh, but figured that as I'm changing it anyway, might as well align it with the rest of the code.

Updated as per your suggestion.

XECDesign:
Just one last thing. There's a stray `export BASE_DIR` here https://github.com/RPi-Distro/pi-gen/blob/97076ab9f849c088e256e0bf6ba9f074d1ec0a7b/build.sh#L183

LCvanDinteren:
Nice catch, my bad!

XECDesign:
Many thanks

LCvanDinteren:
Many thanks to you as well!

PR #349, opened 2019-11-18
Add QCOW2 build mechanism

pandel:
**What's this?**
This PR integrates a build variant utilizing qcow2 based images. It is mainly focused on reducing build time and space. Some additional changes have been made to the export stages for this as well, but are only activated if you use the qcow2 mechanism.

**How does it work?**
This mechanism replaces the copy_previous style build by creating a base qcow2 image in WORK_DIR at the beginning of the first stage and adding additional so called backing images stage by stage on top of it. These backing images record only the changes of the stage they belong to. That leads to reduced space consumption and faster stage creation times. <s>Only the two export stages (image and noobs) are created "the old fashioned" way, because it would need a more in depth rework of both export stages to use qcow2 for them as well. I think that this is not really necessary.</s> Qow2-based export-image stage has been added with with https://github.com/RPi-Distro/pi-gen/pull/349/commits/4ea3675b52c01dcf5227127afed99c44a925a7b9, see my comment below.

**I dont' like it! How to deactivate?**
In this PR, qcow2 based building is activated by default. If you want to activate the normal build variant, set `USE_QCOW2=0` inside your `config` file.

**Further information**
You will find further descriptions inside the updated README.md as well.

_I developed and tested it under Kubuntu 18.04.03 LTS, with and without Docker and it works for me (not taking the current qemu/i386 problem into account, as this is all about image building)._


XECDesign:
Looks alright at a glance, thanks. It might be a while before I can go through it properly.

pandel:
> Looks alright at a glance, thanks. It might be a while before I can go through it properly.

Thank! Take your time!

pandel:
Sorry for the additional commits. I try to stay up to date with the main repo while adding some features and make the whole patch more robust...

Commit https://github.com/RPi-Distro/pi-gen/pull/349/commits/4ea3675b52c01dcf5227127afed99c44a925a7b9 is the biggest so far, because I thought about the export stages again. As you know, when I opened this pull request I thought it wouldn't make any big difference to change the export stages.

But I found out that if I would create the base qcow2 image with the 2-partition layout (vfat/ext4) right from the beginning, I could use simple calls to `qemu-img rebase`, `qemu-img commit` and `qemu-img convert` to build an export image only with the additional amount of space the first stage needs. As a side effect, image building runs a even faster with it, too.

pandel:
For anyone who is wondering about the many syncs: I am running the whole thing from within a Ubuntu 18.04 VM under VMware Workstation on a laptop and on a DebianARM64 VM under Proxmox, respectively. Both systems can get under a heavy load and if the overall load gets too high, the image operations inside the VM fail from time to time. Sync makes it work...

pandel:
Manually merged the last upstream commits, don't know why Github isn't able to automatically merge...

bbinet:
This seems a useful addition to improve speed and space!

Any chance to see it merged in upstream repository?

pandel:
As I am using this in all my derived pi-gen projects and never went back to traditional build mechanism, as it simply is too slow and uses too much space, I can assure you that this speeds up things like hell :-) - I don't understand why there is so little to no reaction but who knows...

I tried to keep up with the main repo the last months, resolve conflicts and stuff, but for now, from my side, I will not fix the mentioned conflicts at the moment, as they are simply of no interest to me.

bbinet:
@XECDesign Any chance this can get merged soon?

XECDesign:
Yeah, it's a worthwhile change, but I can't spend time checking it. Given that there's now verification that it works from another user, I am a bit more comfortable with accepting it as is.

pandel:
If I find time the next days I can add the last changes from master branch to my branch to resolve the conflicts if you wish...

XECDesign:
That would be much appreciated.

koenvervloesem:
I'm currently using pi-gen to build a custom image and the build needs half an hour on GitLab CI. I'm definitely interested in testing this PR if the conflicts have been resolved. That will give you some extra verification, @XECDesign.

@pandel How much of a speedup can I expect from a half-hour build in the traditional way?

pandel:
@koenvervloesem I really can't say. It really depends on the underlying hw or vm you are using, cached debs or not, network speed. If you have a very fast hdd, then it might not be that much, because the file copying mechanism in the "old" ways is fast enough. For me, creating a base qemu and then only derived ones for every next step is much faster.

pandel:
So, all conflicts resolved for now. For anyone who wants it is ready to test and use - simply check out my fork and qcow2 branch, create a single line config file and build it. FWIW, I am building and testing solely with Docker, as I am usually building under Mac OS or unsupported Linux environments. Docker serves as standardised build environment in my case.

@koenvervloesem 
For you, as non-representative speed test :-), here the data from my last build test:
- Laptop Lenovo T530, i5-3380M, 4GB, SSD MX500 1TB, WIFI
- Internet connection 450Mbit/s

-> Start at 0:28 (docker with no container and image)
-> End at 02:09 Done! (all 3 images as zip and noobs folders exported)

So about 1:40 for a standard Raspbian lite+standard+full build. Not that bad, I think.

koenvervloesem:
Thanks @pandel! Today I have run some tests on my build with your qcow2 branch. It's a Raspbian Lite (up to stage2) build with some customizations, so the speedup potential is obviously minimal, but I'm welcoming every minute I can shave off. Compared to the 'traditional' pi-gen way of building, the build time has decreased from 30 minutes to 24 minutes roughly. This looks fine to me!

This is all with the `./build-docker.sh` script, this is the only way I have ever been able to build pi-gen reliably. On my local Ubuntu 19.10 machine this works, but in GitLab CI this fails already in stage 0:

```
[16:18:06] Begin /pi-gen
[16:18:06] Begin /pi-gen/stage0
modprobe: ERROR: ../libkmod/libkmod.c:586 kmod_search_moddep() could not open moddep file '/lib/modules/4.19.78-coreos/modules.dep.bin'
modprobe: FATAL: Module nbd not found in directory /lib/modules/4.19.78-coreos
```

I'm not sure where to start with debugging this, but I'll have to delve into this later.

At least now @XECDesign  has external confirmation that this works in a 'normal' setup.

pandel:
@koenvervloesem A little bit better is still better :-) - as I said, if you have only few stages with a small amount of files, the benefit will not be that much...

And there it comes, Gitlab CI... as I mentioned, I have no experience with it. For the mod to work, you need a kernel that is capable of using network block devices (nbd), as any system working directly with qcow2 files. So, you need kernel module named „nbd“ for this to work.

Are you hosting your own instance of Gitlab (as I do) or are you using the official Gitlab server? A quick search regarding the used CoreOS Kernel had no result regarding the included modules. I think, the module is there, but the question is, if you are allowed to load a kernel module during your build process. Sadly, I have no time at the moment to dive deeper into the documentation to search for a solution. So it might be, that you have to stick to the traditional build for now. Sad sad sad...

koenvervloesem:
I'm using the official GitLab server. But I'm already glad you have saved me five minutes for each local build. Good stuff!

pandel:
@koenvervloesem 
I have an idea. What if the functionality is already in the kernel?

I am loading the nbd module during initialization of the first qcow2 image. You could try to disable this and see, if it just works.

To do this, open scripts/qcow2_handling, find the line with "modprobe nbd max_part=16" and delete it. Then start your build process. If it works, I can try to circumvent the error.

koenvervloesem:
@pandel Good thinking, thanks for your help! Unfortunately this doesn't seem to be the case:

```
[06:32:45] Begin /pi-gen
[06:32:45] Begin /pi-gen/stage0
cat: '/sys/class/block/nbd*/size': No such file or directory
[06:32:45] Unloading image
```
How many nbd devices do you need? Maybe I can create them with `mknod`. I already had to patch the `build-docker.sh` script to create loop devices this way, otherwise pi-gen fails to build on GitLab CI (related: https://github.com/RPi-Distro/pi-gen/issues/320), so maybe this does the trick too.

pandel:
@koenvervloesem 
In theory I only need two (one for each partition: boot+root), but sometimes they tend to "hang" around a bit after releasing, so I am assigning names dynamically in the free range the moment I need them. The "max_part=12" parameter generates 12 possible devices on module load, which is more than enough. Sadly, you cannot simply create nodes, the module has to be loaded for this to work. The moment you load it, the devices will be created.

I have no clue how Gitlab CI configuration and job runners work, but isn't it possible, before you start build-docker.sh in your runner configuration, to try a "sudo modprobe nbd max_part=12", or aren't you given root access on the runner OS?

BTW, how does this runner thing work? From what I read is that you have a yml file that configures what to do, but don't you need a separate (vm) machine where the runner lives inside? So, isn't it possible to use some other OS than CoreOS for that matter?

koenvervloesem:
I tried to modprobe the module before the start of `build-docker.sh but that failed with:

```
ip: can't find device 'nbd'
nbd                    45056  1 
ip: can't find device 'max_part=12'
modprobe: can't change directory to '/lib/modules': No such file or directory
```

I tried a couple of other things, but frankly I'm hesistant to invest much more time in it, because I already had to jump through a lot of hoops to get pi-gen to build on GitLab CI and only after a lot of experimenting I found the magic combination of Docker in Docker and manually created loop files that worked. It's clear that this is a fragile combination :-) I fear it needs some more changes to support your QCOW2 build mechanism, and that will probably need a lot more testing before it converges on another magic combination.

Thanks for you help anyway! I'm already glad I can use the mechanism for my local builds, I'll definitely enable your QCOW2 build mechanism there. But for the builds on GitLab CI I'll stay with the current build method.

pandel:
@koenvervloesem 
Ok, I totally understand! Docker in Docker sounds weird by itself.

But I am still curious... The second line in your codeblock looks like somewhat unusual... Could it be that your CoreOS Docker container just needs /lib/modules from the underlying OS as read-only volume? Then it might find the module it needs. Just speculating.

But nevertheless, fine that my mod helps at least a little :-)

Stay safe these days ;-)

XECDesign:
Everybody happy for this to be merged?

pandel:
"Everybody" is a good question... What about you as maintainer? Did you ever test it yourself?

XECDesign:
Just did and the filesystems it produces seem good. And it shaves half an hour off a clean stage4 image build on my PC, which normally takes 1h45m.

A quick skim through the PR shows that there's a lot of mixed whitespace. Generally tabs are used, but this PR doesn't seem to be consistent.

The image size difference is a bit of a concern:

```
4.1G    qcow.img
3.6G    std.img
```

The partition tables don't match up either - they must be 4M aligned, at least. The original has a bug where the boot partition is off by one M at the end, but that doesn't cause any issues.

pandel:
Many thanks for reporting! So, things to do:

- I have to clean up the whitespaces. Time consuming, but simple...

- Image size is a bit weird, as I am doing nothing else than creating disks. I did not change anything regarding the installation process itself. Although I don't think some more MB is relevant these days and anyway, I prefer the NOOBS/PINN solution over the badly slow and less versatile image copy thing, I'll dive into this to find out whats happening - perhaps it is the structure of the image which leads to less compressability or so...

- That the partition tables don't match is correct. I create a standard partition table via sfdisk, which aligns to 1MB per default. Should be quite easy, I will change this as requested.

I'll try to solve everything as soon as possible.

XECDesign:
I haven't checked, but I guess you're using resize2fs -M to shrink the partition and then adjusting accordingly? I don't know the internals of ext filesystems, but I think a certain percentage of blocks ends up being reserved. The 'stock' approach is to create the image based on the actual used space and then copy the files in, so it's closer to the actual used space.

I don't know how much of an issue it is, but when you're imaging thousands of cards for resale, spending a bit of time writing 0s ends up costing more.

I'm not really sure what a good solution would be. Maybe using the old export method of creating an image rather than converting the qcow image? It looks like shrinking the qcow image and converting it to qcow2 takes a bit of time as well. The difference is about 2 minutes (cuts 6 down to 4). So it doesn't seem like a huge sacrifice.

pandel:
Ok, first edits done, whitespace cleaned (not pushed yet :-) ). And a quick look at my resize function revealed that I added an extra amount of space as root margin. I simply remove that in my next test and we'll see how it comes out. Writing 0s is clearly Sisyphos work...

pandel:
Pushed my changes:

- whitespace clean up

- partition alignment corrected, please check

- files size reduced: stage4 build on my Mac uses 3512M inside zip

Please re-test :-)

XECDesign:
Image size looks okay, but alignment doesn't seem right - 4 sectors rather than 4MiB.

XECDesign:
Partition id should be 0c, not 0b. Not sure if it makes a difference, but I remember having obscure issues in some cases, so I'd like to keep changed to a minimum.

pandel:
Many thanks! I look into it as soon as possible. The change should alteady have corrected alignment but I will find the reason why it didn‘t!

pandel:
Sorry, sorry, sorry! No time for nothing...

Alignment and partition type should be correct now. I also added the bootable flag to the FAT partition, just in case.

Please re-test a last time, hopefully...

andrum99:
> I also added the bootable flag to the FAT partition, just in case.

The bootable flag is ignored on the Raspberry Pi.

pandel:
> > I also added the bootable flag to the FAT partition, just in case.
> 
> The bootable flag is ignored on the Raspberry Pi.

As I said, just in case ;-) - and it doesn't hurt anyway.

pandel:
@XECDesign My last tests are working fine. Any recommendations? Is the partition layout now right for you?

pandel:
@XECDesign Any news regarding this PR? Does it work for you?

XECDesign:
Thank you for keeping this up to date and sorry about delaying this again. One more bug release image release and then I'll circle around to PRs again.

XECDesign:
I'll just merge this now to avoid further delays and address any issues with it if they come up.

Thanks

PR #351, opened 2019-11-26
Explicitly check if postrun script exists in BASE_DIR

LCvanDinteren:
This ensures execution of the postrun script, even when any of the previous scripts didn't end in the base directory.


XECDesign:
Makes sense. Thanks

PR #354, opened 2019-11-26
Remove unreleased note

RaspberryPiFan:


XECDesign:
The UNRELEASED note is a necessary part of the build process.

https://github.com/RPi-Distro/pi-gen/blob/master/export-noobs/00-release/00-run.sh#L40

PR #356, opened 2019-11-29
Attach to bridge network

drwilco:
To allow docker-build.sh access to the apt-cacher container, attach to the default network.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,7 @@ version: '2'
 services:
   apt-cacher-ng:
     restart: unless-stopped
+    network_mode: "bridge"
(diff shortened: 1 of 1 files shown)


samtygier:
I confirm this is needed to use the apt-cacher from docker. I have the same fix in my local repo, but had not made a PR yet as I had not tested with the non docker build.

In the docker case, the I find that the ip of the proxy should be

`APT_PROXY=http://172.17.0.2:3142`

Is that also true if run without the build in docker? Eitherway the README could be updated to mention it.

samtygier:
For the non docker build either .1 or .2 work.

PR #359, opened 2019-12-06
Updated info about stage4 and stage5

RaspberryPiFan:


andrum99:
You've removed some things there. If you're going to do that, what about using saying that Stage 4 is used to build the "Raspbian Buster with desktop" image, and Stage 5 is for "Raspbian Buster with desktop and recommended software"?

RaspberryPiFan:
@andrum99 I already wrote it's Raspbian and Raspbian Full, I think that's enough.

XECDesign:
Looks okay, thanks.

PR #362, opened 2019-12-12
move picraft to stage where minecraft is installed

chrisdecker1201:
Hello,

I'm not really sure about that, but for my understanding it's only necassary to install python-picraft when there is a installed minecraft.


XECDesign:
Makes sense. Thanks

Issue #363, opened 2019-12-20
Too simple WPA_PASSWORD aborts script


julled:
Hi,

if i try to use WPA_PASSWORD = "AAA"  or "" as password, the script aborts without any further hints that this might be a problem:

> 
> 'files/wpa_supplicant.conf' -> '/pi-gen/work/2019-12-20-piImage/stage2/rootfs/etc/wpa_supplicant/wpa_supplicant.conf'
> 
> real	0m34,597s
> user	0m0,029s
> sys	0m0,037s

I searched for a while to find this issue.
Maybe a hint in the readme could help against this.

samtygier:
So there are a couple of issues here.
1) wpa_passphrase writes its error messages to stdout, so they get written to wpa_supplicant.conf instead of being visible. We could use tee so that it is visible.
2) No early check. It would be simple to check this when reading variables rather than wait for it to reach stage2

I'll prep some patches for each.

samtygier:
> wpa_passphrase writes its error messages to stdout

This is now fixed upstream, but it would be good to fix it here as it will take a while for the change to reach the raspbian repos. https://w1.fi/cgit/hostap/commit/?id=32289112c40ba3f6d0b3c04263654d6223055e22

XECDesign:
Thank you both.

PR #366, opened 2019-12-23
Handle WPA_PASSWORD length issues

samtygier:
Fixes for #363

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -220,6 +220,11 @@ if [[ -n "${APT_PROXY}" ]] && ! curl --silent "${APT_PROXY}" >/dev/null ; then
 	exit 1
 fi
 
+if [[ ${#WPA_PASSWORD} -lt 8 || ${#WPA_PASSWORD} -gt 63  ]] ; then
+	echo "WPA_PASSWORD" must be between 8 and 63 characters
+	exit 1
+fi
+
 mkdir -p "${WORK_DIR}"
 log "Begin ${BASE_DIR}"
 
--- a/stage2/02-net-tweaks/01-run.sh
+++ b/stage2/02-net-tweaks/01-run.sh
@@ -12,7 +12,8 @@ fi
 
 if [ -v WPA_ESSID ] && [ -v WPA_PASSWORD ]; then
 on_chroot <<EOF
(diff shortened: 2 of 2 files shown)


build.sh
```diff
@@ -220,6 +220,11 @@ if [[ -n "${APT_PROXY}" ]] && ! curl --silent "${APT_PROXY}" >/dev/null ; then
 	exit 1
 fi
 
+if [[ ${#WPA_PASSWORD} -lt 8 || ${#WPA_PASSWORD} -gt 63  ]] ; then
```
XECDesign: Is this length check valid? Can the passphrase have escape characters, for example?

samtygier: From wikipedia the WPA pass-phrase should be"8 to 63 printable ASCII characters". https://w1.fi/cgit/hostap/tree/wpa_supplicant/wpa_passphrase.c checks that there are no control characters.

With this patch if the user does something tricksy like

`WPA_PASSWORD="abcd"$'\n'"efgh"`

Then it will pass this test (9 chars) and fail later on in wpa_passphrase, but at least the error message will be visible.

Options would be:

1) No check here, just let wpa_passphrase catch the  errors (other half of patch makes that visable).

2) Partial check/current patch. Just check length

3) Full check. Try to catch all errors early.

stage2/02-net-tweaks/01-run.sh
```diff
@@ -12,7 +12,8 @@ fi
 
 if [ -v WPA_ESSID ] && [ -v WPA_PASSWORD ]; then
 on_chroot <<EOF
-wpa_passphrase "${WPA_ESSID}" "${WPA_PASSWORD}" >> "/etc/wpa_supplicant/wpa_supplicant.conf"
+set -o pipefail
+wpa_passphrase "${WPA_ESSID}" "${WPA_PASSWORD}" | tee -a "/etc/wpa_supplicant/wpa_supplicant.conf"
```
XECDesign: Don't both of these behave exactly the same way?

```
$ false >> /dev/null; echo $?
1
```

XECDesign: Ah, I see the problem. wpa_passphrase prints everything to stdout...

XECDesign:
Thanks

Issue #368, opened 2020-01-07
WPA_PASSWORD unset


kraeml:
I run into an error "WPA_PASSWORD" must be between 8 and 63 characters.

I don't want a WPA_PASSWORD because I use my wifi setup by own scripts.

Workaround is to set WPA_PASSWORD=dummy_password AND no WPA_ESSID. No errors but not so elegant. If no WPA_PASSWORD is set then no length test is required.

May look at:

https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash

samtygier:
My bad sorry, I didn't check the case of WPA_PASSWORD being unset in #366 . I'll prep a new patch.

PR #370, opened 2020-01-07
Fix missing information in README.md

snwfdhmp:
After following README.md instructions, I see the script failing because I didn't set WPA_PASSWORD:

```
$ sudo ./build.sh
WPA_PASSWORD must be between 8 and 63 characters
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ The following environment variables are supported:
 
    Password for the first user
 
- * `WPA_ESSID`, `WPA_PASSWORD` and `WPA_COUNTRY` (Default: unset)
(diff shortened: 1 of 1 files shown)


samtygier:
`WPA_PASSWORD` shouldn't actually be required, its an error that the length check happens with `WPA_PASSWORD` unset,  #368

andrum99:
I've opened #371 with the correct fix.

snwfdhmp:
Ok, I though forcing WPA_PASSWORD to be set even for Rpi that doesn't support WPA (no wifi) is a good practice to enforce security.

andrum99:
It is certainly good practice to use a password, but the ability to have no password was removed by mistake.

XECDesign:
Why would pre-configuring a wlan connection be good practice? You're giving the passphrase to everybody who has access to the image.

PR #376, opened 2020-01-16
Fix hostname leakage from build host

cshoredaniel:
Many hosts (including the docker host) set HOSTNAME in the environment,
which means that unless the HOSTNAME is explicitly set in a 'config'
that the resulting image hostname will be the build host's hostname.

We fix that by unsetting HOSTNAME prior to reading the configs and
applying defaults.

Signed-off-by: Daniel F. Dickinson <cshored@thecshore.com>

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -125,6 +125,10 @@ fi
 BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 export BASE_DIR
 
+# Make sure HOSTNAME doesn't leak in from host
+unset HOSTNAME
+export HOSTNAME
+
 if [ -f config ]; then
 	# shellcheck disable=SC1091
 	source config


bit-bash:
Oh I discovered this too, had to force set HOSTNAME in config.  Since build.sh contains the line:
export HOSTNAME=${HOSTNAME:-raspberrypi}
it's trying to default HOSTNAME=raspberrypi if HOSTNAME is not defined.

I like your idea for the fix, but I think it would not allow something like this to work:
HOSTNAME=nirvana ./build.sh
Not sure if this is what was intended (I certainly don't do this), I'll either rely on the default or change HOSTNAME in config.

XECDesign:
Strange. I have a clear recollection of fixing this already, but can't find any evidence of that. My imaginary fix was to rename the parameter to something like "GUEST_HOSTNAME"

cshoredaniel:
Ironically I decided against a PIGEN_HOSTNAME solution because I was worried about backward compatibility.  I didn't realize about the HOSTNAME=george ./build.sh and similar use case, and renaming the variable seems the only solution that will work for that case but not end up with host environment HOSTNAME.

XECDesign:
Ah.... found it:
https://github.com/RPi-Distro/pi-gen/commit/24bb01dec280f568ac59cdd1253a33a6d30ac4c4

cshoredaniel:
I think this can be safely closed.

PR #380, opened 2020-01-18
Enable adding SSH pubkey and setting pubkey only

cshoredaniel:
This commit add the ability to specify an SSH public key as well as the
option to disable password authentication and only allow public key
authentication for SSH.

Signed-off-by: Daniel F. Dickinson <cshored@thecshore.com>


ericmigi:
this would be super useful! bump :)

ilyas28:
Could you merge this, as it is already approved?

XECDesign:
Thanks for the nudge.

rvbuelow:
You're welcome 😄

rvbuelow:
You're welcome 😄

PR #384, opened 2020-02-12
BusyBox realpath does not support -s flag, fallback to use no flags

timvisee:
Not all `realpath` implementations support the used `-s` flag, such as the BusyBox version.

I'm using this on an Alpine based host which uses BusyBox. This patch makes the `realpath` command fall back to omit the `-s` flag on error.


XECDesign:
Thanks

Issue #194, opened 2018-06-20
Build fail due to 400 Bad Request


Chan9390:
I found that some of my custom builds fail due to server error during `apt-get install`. 
Here's an example:

```
Get:926 http://mirror.glennmcgurrin.com/raspbian stretch/main armhf vlan armhf 1.9-3.2 [36.1 kB]
Fetched 589 MB in 4min 28s (2190 kB/s)
E: Failed to fetch http://mirror.glennmcgurrin.com/raspbian/pool/main/libx/libxrender/libxrender1_0.9.10-1_armhf.deb  400  Bad request
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

From the `build.sh` script:
https://github.com/RPi-Distro/pi-gen/blob/8d9cfb3e467c83b80c496a2c4c1cb2577eb81ccf/build.sh#L28-L37

The script doesnt check if all the packages were correctly installed.

Chan9390:
I would suggest to add `apt-get install --fix-missing` so that even if some deb file wasnt downloaded due to server error, it will try again once. If the previous command had installed successfully, then the command would not take any action.

PR #387, opened 2020-02-20
Add Fix Missing to APT install

luigifreitas:
This Pull Request adds the ability to `apt-get install` to retry to download a `.deb` file a error happens.


XECDesign:
Isn't `fix` a bit of a misnomer here? I don't see the man page for apt-get saying it will retry the download, just that it will ignore the failure and carry on, holding the missing packages back.

luigifreitas:
Yes, this seems to be the case. I trusted the last comment of #194. I updated the patch with something that actually works. This will try to fetch the asset from the mirror three times before exiting with an error. Thus, making it more resilient against network failures.

XECDesign:
Makes sense. Thank you

PR #391, opened 2020-02-26
Exposed RELEASE as a configurable variable

DragonEagle:
Exposed RELEASE as a configurable variable and updated README.md.  As a
bonus, there is only on place to change the release name when the next
release is stable now.


XECDesign:
Makes sense. Thanks.

PR #393, opened 2020-03-03
Customize names and descriptions of NOOBS bundles

Taikuh:
For those who use NOOBS to view and install their custom-rolled OS.

Now, users can set a `NOOBS_NAME` variable in config. This name will have the appropriate Lite or Full suffix depending on the EXPORT_NOOBS stage.

Descriptions can also be personalized.  `PORT_OF` changes the A port of **Debian** with... text.  `NOOBS_DESCRIPTION_ALL` sets the same description for all EXPORT_NOOBS stage. `NOOBS_DESCRIPTION_#` manually defines the description for a specific stage #.


build.sh
```diff
@@ -192,11 +193,18 @@ export PREV_STAGE_DIR
...
 export IMG_SUFFIX
-export NOOBS_NAME
-export NOOBS_DESCRIPTION
 export EXPORT_DIR
 export EXPORT_ROOTFS_DIR
 
+export NOOBS_NAME="${NOOBS_NAME:-Raspbian}"
+export NOOBS_BASENAME="${NOOBS_NAME}"
+export PORT_OF="${PORT_OF:-Debian}"
+export NOOBS_DESCRIPTION
```
Taikuh: I am not sure if `export NOOBS_DESCRIPTION` is still needed

Taikuh:
I botched my commit squash. I think I went 1 commit too far. I'm sorry I'm still not well-versed with Git. Any way to fix this?

PR #411, opened 2020-05-30
Solve the space in base path problem #75

gromain:
I believe this would solve the root cause of #75. 
However, this does not solve the problem upstream with debootstrap.
Maybe a check should be added so that if there is a space in the base path, debootstrap does not fail silently. Or to recommend using the docked build script. I believe explicit errors are better than undocumented behavior because of an upstream.

If there is a space in the base path anywhere, the scripts fails to start. 
This is due to the default separator used in bash (which is a space). 
This default separator can be change by setting the variable $IFS.
In `build.sh`, the old IFS variable is saved and then replaced by 
`\n\b`. At the end of the script, the saved IFS is set back.

In `build-docker.sh`, a simple double quote was enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # pi-gen
 
-_Tool used to create the raspberrypi.org Raspbian images_
+_Tool used to create the PlantoScop images for the Raspberry Pi_
+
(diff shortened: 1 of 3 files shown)


gromain:
Hey, I just saw that commit 7e1d1c9 has been added by github to this PR. I'll remove it now, it has nothing to do with the problem.

PR #415, opened 2020-06-02
Automagically use i386/debian:buster

aniongithub:
when running on 64-bit host to prevent error #271

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM debian:buster
+ARG BASE_IMAGE=debian:buster
+FROM ${BASE_IMAGE}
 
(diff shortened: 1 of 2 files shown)


rnestler:
Does this really work around the issue? Isn't just the host version of `qemu-arm-static` used anyways?

rnestler:
> Does this really work around the issue? Isn't just the host version of qemu-arm-static used anyways?

I tested it and it seems to work indeed.

cvabc:
It seems with this fix, `blkid` will not show USB disk?

XECDesign:
Sorry about the delay.

PR #417, opened 2020-06-05
Solve the space in base path problem issue #75

gromain:
Closed previous PR #411 and reopening because of mis-management on my side.

I believe this would solve the root cause of #75. 
However, this does not solve the problem upstream with debootstrap.
Maybe a check should be added so that if there is a space in the base path, debootstrap does not fail silently. Or to recommend using the docked build script. I believe explicit errors are better than undocumented behavior because of an upstream.

If there is a space in the base path anywhere, the scripts fails to start. 
This is due to the default separator used in bash (which is a space). 
This default separator can be change by setting the variable $IFS.
In `build.sh`, the old IFS variable is saved and then replaced by 
`\n\b`. At the end of the script, the saved IFS is set back.

In `build-docker.sh`, a simple double quote was enough.

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -41,7 +41,7 @@ if test -z "${CONFIG_FILE}"; then
 	exit 1
 else
 	# shellcheck disable=SC1090
-	source ${CONFIG_FILE}
+	source "${CONFIG_FILE}"
 fi
 
 CONTAINER_NAME=${CONTAINER_NAME:-pigen_work}
--- a/build.sh
+++ b/build.sh
@@ -229,7 +229,7 @@ fi
 mkdir -p "${WORK_DIR}"
 log "Begin ${BASE_DIR}"
 
-SAVEIFS=$IFS
+SAVEIFS="$IFS"
 IFS=$(echo -en "\n\b")
 
 STAGE_LIST=${STAGE_LIST:-"${BASE_DIR}"/stage*}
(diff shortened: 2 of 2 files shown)


gromain:
Any comments on this?

XECDesign:
Shouldn't the variable assignments be double quoted?

gromain:
I just pushed the change you requested (sorry, it fell through the cracks, and I completely forgot about it!).

Is there anything else?

gromain:
This PR creates an issue during the image creation steps (more specifically, line 302 when transforming the collected EXPORT_DIRS array). I'm going to have about ways to solve this.

XECDesign:
It's probably not worth spending too much time on this while debootstrap itself doesn't support spaces. Maybe your suggestion of just flagging up the issue before the build starts is the way to go for now?

gromain:
Yes, I agree. I'll close and recreate a PR with something in the README somewhere, if that's good for you.
Maybe we can put a check too in the build script, to check for the presence of a space in the path and avoid building in that case?

XECDesign:
> Yes, I agree. I'll close and recreate a PR with something in the README somewhere, if that's good for you.

Yeah, sounds great. No need to re-create the PR, you can just modify your existing branch and it will update the PR.

> Maybe we can put a check too in the build script, to check for the presence of a space in the path and avoid building in that case?

Yup, that's what I meant.

Issue #422, opened 2020-06-15
When specifiying FIRST_USER_NAME, it is not added to sudoers correctly


FrancisTurner:
If you specify FIRST_USER_NAME in the config file then almost everything works but when
doing sudo with it, you need to enter the password. This is because the file 010_pi-nopasswd in sudoers.d still references user pi and is not changed to FIRST_USER_NAME

The fix is to add the following to stage2/01-sys-tweaks/01-run.sh around line 47

```
# Add line to replace the username for pi in sudoers.d/010_pi-nopasswd
on_chroot <<EOF
sed "s/pi/$FIRST_USER_NAME/" -i /etc/sudoers.d/010_pi-nopasswd
EOF
# End replace username 
```

PR #423, opened 2020-06-15
Fix issue 422 (stage2/01-sys-tweaks/01-run.sh )

FrancisTurner:
https://github.com/RPi-Distro/pi-gen/issues/422

### DIFF
--- a/stage2/01-sys-tweaks/01-run.sh
+++ b/stage2/01-sys-tweaks/01-run.sh
@@ -45,6 +45,12 @@ for GRP in adm dialout cdrom audio users sudo video games plugdev input gpio spi
 done
 EOF
 
+# Add line to replace the username for pi in sudoers.d/010_pi-nopasswd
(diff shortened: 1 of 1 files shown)


tcv-git:
Hi Francis, this doesn't need to be done in a chroot.  Just use the host sed:

    sed "s/pi/$FIRST_USER_NAME/" -i "${ROOTFS_DIR}/etc/sudoers.d/010_pi-nopasswd"

FrancisTurner:
Good point. I broke it out of another change I made that was using chroot. I'll update the patch

> Hi Francis, this doesn't need to be done in a chroot. Just use the host sed:
> 
> ```
> sed "s/pi/$FIRST_USER_NAME/" -i "${ROOTFS_DIR}/etc/sudoers.d/010_pi-nopasswd"
> ```

tcv-git:
It would be great if they ever merged some of these!

PR #425, opened 2020-06-28
ARM64 support

AaronDewes:
## Changes

- Add ARM64 build flag
- Add feature to add -arm64 or -armhf to a folders name to only run it in that type of build
- Add Bookshelf and Raspberry Pi Diagnostics to make it more similar to the latest Raspberry Pi OS
- Fix to use debootstrap instead of qemu-debootstrap on arm64.
- Use Debian apt repos instead of Raspbian. (I don't know if the current version does this, I don't use the Raspberry Pi OS, but I couldn't find another arm64 repo, should I set it to only do this if arm64 is enabled?)
- Remove omxplayer (https://github.com/popcornmix/omxplayer says it's deprecated)


AaronDewes:
@XECDesign Why did you remove Scratch 3? An ARM64 version is available in your repos: http://archive.raspberrypi.org/debian/pool/main/s/scratch3/ (http://archive.raspberrypi.org/debian/pool/main/s/scratch3/scratch3_1.0.5_arm64.deb)

PR #428, opened 2020-07-17
Replace bsdtar with libarchive-tools

arn4v:
bsdtar has been included in libarchive-tools post Ubuntu 16.04 (Xenial)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get -y update && \
     apt-get -y install --no-install-recommends \
         git vim parted \
         quilt coreutils qemu-user-static debootstrap zerofree zip dosfstools \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apt-get -y update && \
     apt-get -y install --no-install-recommends \
         git vim parted \
         quilt coreutils qemu-user-static debootstrap zerofree zip dosfstools \
-        bsdtar libcap2-bin rsync grep udev xz-utils curl xxd file kmod bc\
+        libarchive-tools libcap2-bin rsync grep udev xz-utils curl xxd file kmod bc\
```
AaronDewes, line 9: The dockerfile uses debian buster, which has bsdtar, so you're breaking the docker version with this

Issue #358, opened 2019-12-05
Docker build LOOP_SET_STATUS64 failed: Resource temporarily unavailable


jtc42:
Hi,

I'm having trouble getting docker builds to work. I've modified export-image/presun.sh to give extra debug information around the issue, resulting in:

```
7509 Syncing disks.
7510 11780: loopdev:      CXT: [0xfffc84c4]: initialize context
7511 11780: loopdev:      CXT: [0xfffc84c4]: init: ignore ioctls
7512 11780: loopdev:      CXT: [0xfffc84c4]: init: loop-control detected 
7513 11780: loopdev:      CXT: [0xfffc84c4]: find_unused requested
7514 11780: loopdev:      CXT: [0xfffc84c4]: using loop-control
7515 11780: loopdev:      CXT: [0xfffc84c4]: loop2 name assigned
7516 11780: loopdev:      CXT: [0xfffc84c4]: find_unused by loop-control [rc=0]
7517 11780: loopdev:      CXT: [0xfffc84c4]: set offset=4194304
7518 11780: loopdev:      CXT: [0xfffc84c4]: set sizelimit=268435968
7519 11780: loopdev:      CXT: [0xfffc84c4]: set backing file=/pi-gen/work/2019-12-05-raspbian-openflexure-buster/export-imag
7520 11780: loopdev:    SETUP: [0xfffc84c4]: device setup requested
7521 11780: loopdev:    SETUP: [0xfffc84c4]: backing file open: OK
7522 11780: loopdev:      CXT: [0xfffc84c4]: open /dev/loop2 [rw]: Success
7523 11780: loopdev:    SETUP: [0xfffc84c4]: device open: OK
7524 11780: loopdev:    SETUP: [0xfffc84c4]: LOOP_SET_FD: OK
7525 11780: loopdev:    SETUP: [0xfffc84c4]: LOOP_SET_STATUS64 failed: Resource temporarily unavailable
7526 11780: loopdev:    SETUP: [0xfffc84c4]: failed [rc=-11]
7527 losetup: /pi-gen/work/2019-12-05-raspbian-buster/export-image/2019-12-05-raspbian-buster-full.img: failed to set up loop device: Resource temporarily unavailable
7528 11780: loopdev:      CXT: [0xfffc84c4]: de-initialize
7529 11780: loopdev:      CXT: [0xfffc84c4]: closing old open fd
7530 11780: loopdev:     ITER: [0xfffc8644]: de-initialize
```

[...]

PR #449, opened 2020-09-22
workaround build failure on AWS M6g instances

kdoren:
Build was failing for me on AWS M6g Arm instance, apparently the faster CPU triggers a race condition. This change (suggested by jtc42 here: RPi-Distro/pi-gen#358 ) introduces a wait/retry loop to work around the failure.


XECDesign:
No objection to the change in general, but I'm not a fan of a retry loop with no timeout or retry limit.

kdoren:
OK, good point, I added a retry counter.

XECDesign:
thank you

Issue #282, opened 2019-05-01
ssh


mikysal78:
howto for create file authorized_keys for root?
howto change default port ssh?

XECDesign:
You can install files like I do here - https://github.com/RPi-Distro/pi-gen/blob/master/stage2/01-sys-tweaks/01-run.sh

And you can drop in patches, like you see here - https://github.com/RPi-Distro/pi-gen/tree/master/stage2/01-sys-tweaks/00-patches

mikysal78:
> You can install files like I do here - https://github.com/RPi-Distro/pi-gen/blob/master/stage2/01-sys-tweaks/01-run.sh
> 
ok, but create in the home FIRST_USER dir .ssh with root don'user.
Because?
install -d  "${ROOTFS_DIR}/home/ninux/.ssh"
install -d  "${ROOTFS_DIR}/root/.ssh"
install -m 600 files/authorized_keys2   "${ROOTFS_DIR}/home/ninux/.ssh/"
install -m 600 files/authorized_keys2   "${ROOTFS_DIR}/root/.ssh/"

> And you can drop in patches, like you see here - https://github.com/RPi-Distro/pi-gen/tree/master/stage2/01-sys-tweaks/00-patches
>
this is ok for chage default port on ssh

Thanks.

XECDesign:
> ok, but create in the home FIRST_USER dir .ssh with root don'user.
> Because?

Pardon?

mikysal78:
ninux@ninuxberry:~ $ ls -la
totale 24
drwxr-xr-x 3 ninux ninux 4096 mag  1 16:39 .
drwxr-xr-x 3 root  root  4096 mag  1 16:23 ..
-rw-r--r-- 1 ninux ninux  220 mag  1 16:23 .bash_logout
-rw-r--r-- 1 ninux ninux 3523 mag  1 16:23 .bashrc
-rw-r--r-- 1 ninux ninux  675 mag  1 16:23 .profile
drwxr-xr-x 2 **root  root**  4096 mag  1 16:39 .ssh
ninux@ninuxberry:~ $

XECDesign:
You can specify the permissions/ownership you want - https://github.com/RPi-Distro/pi-gen/blob/master/stage4/03-magpi/00-run.sh#L14

`man install`

PR #466, opened 2020-11-11
Make default ssh port configurable

KeithYeh:
User can change ssh port by setting `SSH_PORT` in the `config` file

Related issue: #282


build.sh
```diff
@@ -172,6 +172,7 @@ export WPA_PASSWORD
 export WPA_COUNTRY
 export ENABLE_SSH="${ENABLE_SSH:-0}"
 export PUBKEY_ONLY_SSH="${PUBKEY_ONLY_SSH:-0}"
+export SSH_PORT="${SSH_PORT:-22}"
```
XECDesign, line 175: Since the default port is already 22, why fallback to setting this variable to 22? If you do this, the `-n` condition will not be true and you'll modify the default sshd_config even if there's no change necessary. Then users will be bugged about conffile changes any time sshd_config changes in the debian package.

KeithYeh: I'll change to `export SSH_PORT`

PR #475, opened 2020-11-29
Update preinstalled guide to v4

AaronDewes:


AaronDewes:
I'll probably stop contributing here, it seems like all of my PRs are ignored and the same change gets applied later by the maintainer.

PR #480, opened 2020-12-14
Adds configurable limits for xz memory and thread usage

### LINKED ISSUE #479: xz compression process needs memory limit (export-noobs breaks on certain hosts)
When running a build on a high core-count server, I run into an issue where the number of threads that xz attempts to use (56) effectively starves my machine for memory.

The configuration for xz for the number of threads to use is `-T0` and thus the number of threads and the amount of memory is unbounded by the definition given there and there is no way for the user to bound those parameters at run time through any configuration.

I suggest adding at least an `XZ_MEMLIMIT` or similar named configuration item with a fallback of 0 (infinity) which will allow a user to bound the memory taken by the process by configuration. If the config is not defined, the fallback will take place. The call could look like:

`xz -T0 --memlimit ${XZ_MEMLIMIT:-0}`

I also don't see a reason not to expose the number of cores to allow xz to use here as well. A configuration like `XZ_THREADS` could work?

`xz -T${XZ_THREADS:-0} --memlimit ${XZ_MEMLIMT:-0}`

I'll be following up with a pull request soon, but please let me know if I'm not on the right track here with my proposed solution thus far.

bamartin125:
This pull request closes #479.

It adds the following configuration items

`XZ_MEMLIMIT` in order to allow bounding the memory used by xz (generally by reducing the number of threads)
`XZ_THREADS` in order to allow bounding the number of threads spawned by xz


XECDesign:
Sorry for the very late response. Things have changed a bit since this was submitted and NOOBS is no longer supported. That export stage will probably be removed altogether before the Trixie release, so, I'll close this one.

PR #483, opened 2020-12-15
Adds call to mknod if loop dev does not exist after call to `losetup -f`

### LINKED ISSUE #482: losetup fails with docker-in-docker
The scenario is this:

bamartin125:
This pull request closes #482.

A function called `ensure_next_loopdev` is defined which calls `losetup -f` which requests a new loop device to be created if a free one does not exist already and returns the name to the device that should be then available (but is not yet claimed). Then, the function checks that the device can be seen. If not, then a call to `mknod` is made to create the loop device. This works when the container has the `--privileged` flag.

The `ensure_next_loopdev` call is made before each time a call to `losetup ... -f ... <file>` is made to best ensure that when that call is made, a loop device is available. This implementation could fail in cases where there is another process grabbing the requested loop device between calls to `losetup -f` in `ensure_next_loopdev` and `losetup ... -f ... <file>` (race condition). In these cases, the existing retry loop *should* help.


XECDesign:
Many thanks and sorry it was left hanging for this long

PR #484, opened 2020-12-20
Don't check for binfmt_misc on arm platforms

kormoc:
When building on a native ARM platform, binfmt_misc is not
required to be loaded. This change checks the machine type
and if it's a ARM platform, skip the binfmt_misc validation.


AaronDewes (approved):

XECDesign:
Sorry, better late than never. Many thanks.

PR #489, opened 2021-01-10
config file for build.sh

ssfdre38:
the default config file for build.sh to run


XECDesign:
The file left out intentionally. The user provides their own file, which is not handled by git.

PR #498, opened 2021-03-18
DNS as serveraddress?

ZingenmetJan:
Is it possible to use DNS as Server Address?
It would make it much easier for a group of people, who want to use the Rpi stand-alone. Only one person would need to change the IP-address and each groupmember  can log in automatically.


XECDesign:
Sorry, I don't quite understand the question or how the PR relates to official images or pi-gen in general.

PR #505, opened 2021-04-15
Passwords

D-G-L:
Hi, would you pls post the logon passwords on the downloadable image you supplied? Its not the standard raspian login pi/raspberry


XECDesign:
I believe you may have posted this in the wrong place.

PR #521, opened 2021-06-14
Added PIGEN_DOCKER_OPTS to  build-docker.sh

### LINKED ISSUE #520: It is not possible to pass additional options to the docker build
The docker build has no way to pass additional options to the docker run command.  This is necessary when additional private information (like keys) are not in the git repository but are available on the build machine.

Claudenw:
fixes #520

Setting the PIGEN_DOCKER_OPTS envrionment variable with Docker parameters arguments results in those parameters being added to the Docker run command.

For example
```
export PIGEN_DOCKER_OPTS="-v /etc/apt/certs:/etc/apt/certs"
```

results in the local `/etc/apt/certs` directory being visible inside the docker container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,10 @@ fix is to ensure `binfmt-support` is installed on the host machine before
 starting the `./build-docker.sh` script (or using your own docker build
 solution).
 
+### Passing arguments to Docker
+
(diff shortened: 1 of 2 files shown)


Claudenw:
Added documentation as requested

PR #522, opened 2021-06-16
build-docker.sh updated to allow for empty PIGEN_DOCKER_OPTS 

DBongiorno-CG:
Added simple catch for default case in which PIGEN_DOCKER_OPTS is unspecified during docker build


XECDesign:
What impact does this PR have? As far as I can tell, there's no difference between a variable being empty and it being unset, in this case.

DBongiorno-CG:
Hello! Fair point, I guess they are equivalent in that sense. The intent of this PR was to still allow for a default docker build should a user not specify any additional PIGEN_DOCKER_OPTS.

XECDesign:
I guess I don't understand what you mean by 'allow' in this sense. It still works whether or not the user sets the variable or not as is, right?

DBongiorno-CG:
It did not - when I pulled the repo the other day and ran ./build-docker.sh for my project I received the following: 
> ./build-docker.sh: line 90: PIGEN_DOCKER_OPTS: unbound variable

So I added a statement in that script to set PIGEN_DOCKER_OPTS to a value of "" if unspecified by the user.

XECDesign:
Oh right, look like the author of this script passes the `-u` argument which results in this behaviour. Makes perfect sense, thank you.

PR #526, opened 2021-06-20
Fix: update dependencies

miccies3:
I've been trying to run `pi-gen` on Ubuntu-20.04 and encountered problem while installing dependencies. The tool `bsdtar` is a part of  `libarchive-tools`. I fixed the dependency name where possible. Unfortunately I don't have opportunity to test if the need to switch packagage is or will be necessary on "supported" systems.


XECDesign:
Many thanks

PR #527, opened 2021-06-21
Create config-template

fisiel:


XECDesign:
Although useful, it seems like yet another place that needs to be updated when an option is added/changed. I think the README covers the options well enough.

miccies3:
I agree that README covers it fully up, I just thought, that it would be nice to have already prepared template with docs inside, something similar to config.txt in Raspberry Pi OS

PR #529, opened 2021-06-23
Fix broken imagetool.sh --mount

bbinet:
This pull request fix `imagetool.sh --mount` command which was not working.


XECDesign:
Thanks

PR #531, opened 2021-07-09
Update to use sd8887-nxp package

alexgg:
Replace the proprietary sd8887-mrvl package with the GPL licensed sd8887-nxp dkms source package hosted in cloudsmith.


alexgg:
Wrong origin

Issue #550, opened 2021-10-01
/dev/mapper/nbd10p1: No such file or directory


1000283:
Host: Ubuntu Xenial x64 VM.

I'm trying to build a full image (no SKIPs anywhere) just to test and failing right at stage0.
Using release `2021-05-07-raspbian-buster` and `build.sh`. Same issue on `master`. 

```
$ uname -a
Linux myuser-VirtualBox 4.15.0-129-generic #132~16.04.1-Ubuntu SMP Wed Dec 16 06:46:04 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```
```
$ cat raspbian-full.cfg 
IMG_NAME="Raspbian"
RELEASE="buster"
DEPLOY_ZIP=0
LOCALE_DEFAULT="pt_PT.UTF-8"
KEYBOARD_KEYMAP="pt"
IKEYBOARD_LAYOUT="Portuguese"
TIMEZONE_DEFAULT="Europe/Lisbon"
FIRST_USER_NAME="myuser"
FIRST_USER_PASS="myuser"
ENABLE_SSH=1
```
```
$ sudo CLEAN=1 ./build.sh -c raspbian-full.cfg 
[10:41:44] Begin /home/myuser/dev/pi-gen
[10:41:44] Begin /home/myuser/dev/pi-gen/stage0
Creating base image: image-stage0.qcow2
Formatting 'image-stage0.qcow2', fmt=qcow2 size=12884901888 encryption=off cluster_size=65536 lazy_refcounts=off refcount_bits=16
Checking that no-one is using this disk right now ... OK

Disk /dev/nbd10: 12 GiB, 12884901888 bytes, 25165824 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes

>>> Created a new DOS disklabel with disk identifier 0xe2ad8441.
Created a new partition 1 of type 'W95 FAT32 (LBA)' and of size 250 MiB.
/dev/nbd10p2: Created a new partition 2 of type 'Linux' and of size 11,8 GiB.
/dev/nbd10p3: 
New situation:

Device       Boot  Start      End  Sectors  Size Id Type
/dev/nbd10p1 *      8192   520191   512000  250M  c W95 FAT32 (LBA)
/dev/nbd10p2      520192 25165823 24645632 11,8G 83 Linux

The partition table has been altered.
[...]

PR #552, opened 2021-10-05
Use sync mode when running 'kpartx -a'

Josh798:
This is regarding #550. This enables sync mode whenever we run `kpartx -a`. This ensures that `/dev/mapper/nbdXXpX`actually exists before we try to use it. I think more recent versions of `kpartx` enable sync mode by default, but it seems the `kpartx` that gets installed in Ubuntu Xenial does not.


XECDesign:
Thank you

PR #581, opened 2022-01-19
Fix docker build with debian bullseye image

simonliedl:
- Fix missing sfdisk which is not included anymore in bullseye container
- Fix problem with binfmt_misc not being mounted inside docker although present on host

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -98,6 +98,8 @@ if [ "${CONTAINER_EXISTS}" != "" ]; then
 		--volumes-from="${CONTAINER_NAME}" --name "${CONTAINER_NAME}_cont" \
 		pi-gen \
 		bash -e -o pipefail -c "dpkg-reconfigure qemu-user-static &&
+	# binfmt_misc is sometimes not mounted with debian bullseye image
+	(mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc || true) &&
 	cd /pi-gen; ./build.sh ${BUILD_OPTS} &&
 	rsync -av work/*/build.log deploy/" &
 	wait "$!"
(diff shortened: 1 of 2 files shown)


build-docker.sh
```diff
@@ -98,6 +98,8 @@ if [ "${CONTAINER_EXISTS}" != "" ]; then
 		--volumes-from="${CONTAINER_NAME}" --name "${CONTAINER_NAME}_cont" \
 		pi-gen \
 		bash -e -o pipefail -c "dpkg-reconfigure qemu-user-static &&
+	# binfmt_misc is sometimes not mounted with debian bullseye image
+	(mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc || true) &&
```
XECDesign, line 102: Instead of repeating the same line twice, wouldn't it be doing the same thing just before the if statement?

XECDesign: Ah got it, thanks.

XECDesign:
The repo is frozen until the next release comes out, but I'll merge this once that goes out. Thanks again.

XECDesign:
Many thanks

build-docker.sh
tazle: The order here is wrong. Mounting binfmt_misc needs to happen before dpkg-reconfigure of qemu-usr-static.

bodiroga: Hi @XECDesign! I can confirm what @tazle said: after changing the order of the commands, the error "Module binfmt_misc not loaded in host" has disappeared in my machine (Ubuntu 20.04 on a XPS-13-9310) 😉 This is what I have right now:

```
bash -e -o pipefail -c "(mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc || true) && dpkg-reconfigure qemu-user-static &&
	# binfmt_misc is sometimes not mounted with debian bullseye image
	cd /pi-gen; ./build.sh ${BUILD_OPTS} &&
	rsync -av work/*/build.log deploy/" &
	wait "$!"
```

@tazle, are you willing to provide a PR?

XECDesign: Thanks for testing. Yeah, a PR would be appreciated.

tazle: Sorry, I don't have a current version around, and can't test the whole build process because of #624.

PR #584, opened 2022-02-02
Add check and warning for a space presence in the base path

### LINKED ISSUE #75: Paths with spaces do not work

gromain:
Reopening of #417 from a different upstream repo to keep things cleaner.

This PR closes #75 with a Upstream issue reason. For now, debootstrap doesn't support spaces in path, so we just check for the presence of a space character in the base path, warn the user and exit early before trying anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,24 @@ qemu-utils kpartx
 The file `depends` contains a list of tools needed.  The format of this
 package is `<tool>[:<debian-package>]`.
 
+## Getting started with building your images
+
+Getting started is as simple as cloning this repository on your build machine. You
+can do so with:
+
+```bash
+git clone -–depth 1 https://github.com/RPI-Distro/pi-gen.git
+```
+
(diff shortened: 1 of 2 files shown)


gromain:
Shellcheck is complaining about a few word-splitting issues, do you want me to fix them here too or should I create another PR to keep things clean?

Also, I'm happy to make changes to wording and such in the README and in the warning message.

XECDesign:
Many thanks

XECDesign:
> Shellcheck is complaining about a few word-splitting issues, do you want me to fix them here too or should I create another PR to keep things clean?

Generally separate PRs for separate issues are better.

PR #589, opened 2022-02-16
README contains incorrect dash

jnnks:
git will complain about the malformed dash: `error: unknown non-ascii option in string: -–depth` as there are two different kinds of dashes in the option.


XECDesign:
Oof, well spotted. Thanks!

PR #592, opened 2022-02-17
Addition of a `DEPLOY_COMPRESSION` option to the configuration

gromain:
This PR creates a `DEPLOY_GZ` option available to use in the config file.
This is turned off by default, and will only work if `DEPLOY_ZIP=0`.

As it says on the tin, this addition provides a way to produce compressed images in a gzip format.

`pigz` is used to create the image as it allows the computation to be made using all the available cores, leading to a faster compression stage. It has thus been added to `depends`.


XECDesign:
There has also been talk of switching the official images over to .xz. How about a DEPLOY_COMPRESSION=gz/xz/zip/none type of parameter to avoid having multiple DEPLOY_$format variables in the future?

gromain:
Oh, that would make a lot of sense!
Is there someone working on it? Do you want me to give it a go?

gromain:
I'm doing the latest tests now, but it seems all is working well, with `gzip`, `xz`, `zip` and `none` being the available images format supported.

gromain:
All good on my side for this PR.
Build works perfectly with every settings (`gzip`, `xz`, `zip` and `none`, with `zip` being the default).

export-image/04-finalise/01-run.sh
```diff
@@ -95,11 +95,23 @@ else
 	make_bootable_image "${STAGE_WORK_DIR}/${IMG_FILENAME}${IMG_SUFFIX}.qcow2" "$IMG_FILE"
 fi
 
-if [ "${DEPLOY_ZIP}" == "1" ]; then
+case "${DEPLOY_COMPRESSION}" in
+zip)
 	pushd "${STAGE_WORK_DIR}" > /dev/null
-	zip "${DEPLOY_DIR}/${ZIP_FILENAME}${IMG_SUFFIX}.zip" \
+	zip "${DEPLOY_DIR}/${ARCHIVE_FILENAME}${IMG_SUFFIX}.zip" \
```
XECDesign, line 101: If  we're using compression level 9 for gz and xz, it would probably make sense for zip to do so as well. I suspect people will want this to be configurable though.

export-image/04-finalise/01-run.sh
```diff
@@ -95,11 +95,23 @@ else
...
 		"$(basename "${IMG_FILE}")"
 	popd > /dev/null
-else
+	;;
+gz)
+	cp "$IMG_FILE" "$DEPLOY_DIR/${ARCHIVE_FILENAME}${IMG_SUFFIX}.img"
+	pigz --force -9 "${DEPLOY_DIR}/${ARCHIVE_FILENAME}${IMG_SUFFIX}.img"
```
XECDesign, line 107: A bit of unnecessary IO copying the whole file like that. Why not `-c > "${DEPLOY_DIR}/${ARCHIVE_FILENAME}${IMG_SUFFIX}.img.gz"` (same for xz)?

export-image/04-finalise/01-run.sh
gromain: Makes sense, I'll create an option for this.

depends
```diff
@@ -20,3 +20,4 @@ bc
 qemu-nbd:qemu-utils
 kpartx
 gpg
+pigz
```
XECDesign, line 23: The docker side of things will need this too.

export-noobs/prerun.sh
```diff
@@ -3,10 +3,10 @@
 NOOBS_DIR="${STAGE_WORK_DIR}/${IMG_NAME}${IMG_SUFFIX}"
 mkdir -p "${STAGE_WORK_DIR}"
 
-if [ "${DEPLOY_ZIP}" == "1" ]; then
-	IMG_FILE="${WORK_DIR}/export-image/${IMG_FILENAME}${IMG_SUFFIX}.img"
-else
+if [ "${DEPLOY_COMPRESSION}" == "none" ]; then
 	IMG_FILE="${DEPLOY_DIR}/${IMG_FILENAME}${IMG_SUFFIX}.img"
+else
+	IMG_FILE="${WORK_DIR}/export-image/${IMG_FILENAME}${IMG_SUFFIX}.img"
```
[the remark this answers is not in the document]
gromain: OK

export-image/04-finalise/01-run.sh
gromain: For some reason I felt it would be more efficient this way, but I didn't do any tests to confirm this or not.
You're right though, it makes more sense, I'll fix this. Same for xz I think. I'll do a pass on all of this to reduce unnecessary IO.

README.md
```diff
@@ -116,9 +116,12 @@ The following environment variables are supported:
 
    Output directory for target system images and NOOBS bundles.
 
- * `DEPLOY_ZIP` (Default: `1`)
```
XECDesign, line 119: I wonder if we should preserve backward compatibility and default to `DEPLOY_COMPRESSION=none` if `DEPLOY_ZIP` is '0'. Would be nice not to break things for people who were using the previous variable.

gromain: I was thinking about this too... It makes sense too.

So I'll add that if `DEPLOY_ZIP=0`, we force `DEPLOY_COMPRESSION=none`, but is there anything I should do for `DEPLOY_ZIP=1` too? I'm thinking about the case were someone wrote it explicitly in their config file (admittedly a rare occurrence, since it was the default).
Also, if both are present (`DEPLOY_ZIP=0` and `DEPLOY_COMPRESSION=somethingotherthannone`)for some reason, which one takes precedence? I'm thinking I could print a warning for the user, and make `DEPLOY_COMPRESSION` take precedence.

export-noobs/prerun.sh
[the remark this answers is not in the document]
gromain: (I was following the previous behavior, but this can be changed easily, I agree with your take on this).

README.md
XECDesign: As you say, `DEPLOY_COMPRESSION` should take precedence and `DEPLOY_ZIP` should be marked as deprecated in the documentation.

> is there anything I should do for `DEPLOY_ZIP=1` too?

probably not

export-noobs/prerun.sh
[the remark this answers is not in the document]
XECDesign: Looks like that was snuck in in an unrelated PR which was too large to review properly (lessons were learned from that one).

https://github.com/RPi-Distro/pi-gen/pull/349/files#diff-8636f4b41bccf396c5b46f2eb759a0e2792fc785db5c3fcefb2a6e7bca59d538L97

README.md
gromain: OK.
In the README, do you still want an entry for `DEPLOY_ZIP` with the deprecation notice or just a note in the `DEPLOY_COMPRESSION` section?

I print a warning for the user also on the detection of `DEPLOY_ZIP=0`:
```
DEPLOY_ZIP has been deprecated in favor of DEPLOY_COMPRESSION
Similar behavior to DEPLOY_ZIP=0 can be obtained with DEPLOY_COMPRESSION=none
Please update your config file
```

export-noobs/prerun.sh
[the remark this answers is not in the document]
gromain: OK, removing then! :-)

README.md
XECDesign: > In the README, do you still want an entry for `DEPLOY_ZIP` with the deprecation notice

Yeah, please

export-image/04-finalise/01-run.sh
gromain: Should the default be set to level 9 by the way ?

README.md
gromain: No worries

export-image/04-finalise/01-run.sh
XECDesign: Not sure that I have a strong opinion on this. One one hand, it might take quite a while without much difference in size, on the other hand, it's the sort of thing that runs as a nightly cronjob, so it doesn't matter how long it takes.

I think I'd prefer that the tools used whatever their defaults are.

gromain: OK, they all default to  6, I'll change this then.

export-image/04-finalise/01-run.sh
```diff
@@ -95,11 +95,22 @@ else
...
-	mv "$IMG_FILE" "$DEPLOY_DIR/"
-fi
+	;;
+gz)
+	pigz --force -"${COMPRESSION_LEVEL}" "$IMG_FILE" -c > \
+	"${DEPLOY_DIR}/${ARCHIVE_FILENAME}${IMG_SUFFIX}.img.gz"
+	;;
+xz)
+	xz --compress --force --threads 0 --memlimit-compress=50% -"${COMPRESSION_LEVEL}" \
+	--stdout "$IMG_FILE" > "${DEPLOY_DIR}/${ARCHIVE_FILENAME}${IMG_SUFFIX}.img"
```
XECDesign, line 111: `.xz`?

gromain: Ah yes, my bad! A sad copy paste! :-)

XECDesign:
I'll test it a little bit when I get a chance, just to double check everything, but I think it's good to go. Thanks for your work on this.

gromain:
No worries, I think it's good to battle test it before creating any problems!
And my pleasure, thanks for your work on pi-gen! ;-)

XECDesign:
Many thanks!

PR #599, opened 2022-02-22
Add more binfmt errors and hints to README

jnnks:
Seems like I was not the only one stumbling upon the binfmt errors. Here is a short addition to the readme


XECDesign:
Thank you

PR #602, opened 2022-03-01
Apt proxy fallback

umutediz:
Apt proxy fallback option added as a config file parameter.
If configured to use direct connection as a fallback, the script will test connectivity with the proxy server each time before using it via a bash function for debootstrap and via apt.conf.d/52fallback file on chroot.

If fallback is configured, the old 51cache file will not be installed.

export-image stage removes all installed files related to this functionality. 

Testing this script on Debian buster with the following config file in 4 different scenarios worked fine for me. I tried it while the proxy is down all the time, up all the time, changed state from up to down and down to up during build. Script used proxy whenever it is available in all of these scenarios.

`IMG_NAME='ProxyTest'  
APT_PROXY=http://10.0.2.3:3142/  
APT_PROXY_FALLBACK=1`


umutediz:
Failed to link the issue related with this PR: #602

build.sh
```diff
@@ -280,9 +281,11 @@ if [[ ! "$FIRST_USER_NAME" =~ ^[a-z][-a-z0-9_]*$ ]]; then
 	exit 1
 fi
 
-if [[ -n "${APT_PROXY}" ]] && ! curl --silent "${APT_PROXY}" >/dev/null ; then
-	echo "Could not reach APT_PROXY server: ${APT_PROXY}"
-	exit 1
+if [[ -n "${APT_PROXY}" && ! proxy_check ]]; then
```
XECDesign, line 284: `proxy_check` already has `[[ -n "${APT_PROXY}" ]]`?

XECDesign:
At a glance, looks good. Might be a while before I can go over it fully, test and merge.

build.sh
umutediz: You are right. Early implementations of `proxy_check` were not checking the environment and I forgot to update `build.sh`, probably. Since it is not breaking things, didn't notice it during tests.

umutediz: You are right. Early implementations of `proxy_check` were not checking the environment and I forgot to update `build.sh`, probably. Since it is not breaking things, didn't notice it during tests.

scripts/common
```diff
@@ -7,13 +7,16 @@ bootstrap(){
 	local BOOTSTRAP_CMD=debootstrap
 	local BOOTSTRAP_ARGS=()
 
-	export http_proxy=${APT_PROXY}
-
+	if proxy_check; then
```
XECDesign, line 10: Unless I'm missing something, if proxy_check fails, this seems to be disabling it regardless of whether APT_PROXY_FALLBACK is set.

umutediz: Since bootstrap is one of the earliest operations that fetch files, checking proxy in build.sh shadowed the bug in my test environment, I think. Fixed now.

PR #626, opened 2022-08-10
Fix do not try to count mounts /proc /dev etc.

lukavia:
On a busy server /proc may be to dynamic and du leads to errors like: 
du: cannot access work/stage2/rootfs/proc/4099/task/4912/fd/422'“: No such file or directory
That may lead to wrong size that is too big for truncate later on.

### DIFF
--- a/export-image/prerun.sh
+++ b/export-image/prerun.sh
@@ -11,7 +11,7 @@ if [ "${NO_PRERUN_QCOW2}" = "0" ]; then
 	mkdir -p "${ROOTFS_DIR}"
 
 	BOOT_SIZE="$((256 * 1024 * 1024))"
-	ROOT_SIZE=$(du --apparent-size -s "${EXPORT_ROOTFS_DIR}" --exclude var/cache/apt/archives --exclude boot --block-size=1 | cut -f 1)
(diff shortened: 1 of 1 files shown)


XECDesign:
Under what circumstances would work/stage2/rootfs/proc be mounted at this point? It would indicate that something went wrong during stage2.

Or am I missing something?

lukavia:
I'm sorry. On a server we use for some reason those mounts ware left for every stage. I've push the fix before realizing what would be the case. 
It actually made the things much worse since when the script tries to cleanup up it is a root user trying to delete everything in proc and makes the server unusable and needing restart. 
Unfortunately I couldn't locate where is the problem and why it doesn't unmount those. 
And cannot make more test on this server. I also wasn't been able to reproduce the problem on another machine. 

Although the pull request is still valid you can discard it.

XECDesign:
> It actually made the things much worse since when the script tries to cleanup up it is a root user trying to delete everything in proc and makes the server unusable and needing restart.

Ah yeah, I've definitely been there.

> Unfortunately I couldn't locate where is the problem and why it doesn't unmount those.

Alright, I'll close this, but if you find out the root cause of the issue, then we can add a fix.

PR #636, opened 2022-10-03
Update README.md document WPA_COUNTRY Code format

kralo:
It took me half a day to figure out why the image-building process stopped.

```
  if ! grep -q "^${REGDOMAIN}[[:space:]]" /usr/share/zoneinfo/iso3166.tab; then
    if [ "$INTERACTIVE" = True ]; then
      whiptail --msgbox "$REGDOMAIN is not a valid ISO/IEC 3166-1 alpha2 code" 20 60
    fi
    return 1
  fi
```

raspi-config will silently fail, if the argument to `do_wifi_country` is not an uppercase 2-letter Code from the list in `/usr/share/zoneinfo/iso3166.tab`

Document this in the readme.


XECDesign:
Thanks!

Issue #639, opened 2022-10-20
mkdosfs erroring due to losetup -o incorrectly creating partitions


rfc6919:
running pi-gen on a `5.18.0-0.deb11.4-arm64` system (Debian bullseye+backports, on non-RPi arm64 hardware), the losetup invocation for `BOOT_DEV` in `export-image/prerun.sh` is for some reason creating `/dev/loop0p{1,2}` even though `-P` isn't being passed to losetup. As a result, mkdosfs errors out like so:

`mkdosfs: Partitions or virtual mappings on device '/dev/loop0', not making filesystem (use -I to override)`

I'm not sure why the partitioned loop device is being created (this is presumably not a pi-gen issue), but is there any particular reason the image creation uses 2 loop devices with `-o` and `--sizelimit` specified rather than just a single partitioned loop device?

invocation, running from a checkout of `2bc1f64` on the `origin/arm64` branch. no `config` is present:

```
sudo IMG_NAME=rpi-test-lite STAGE_LIST="stage0 stage1 stage2" DEPLOY_COMPRESSION=none ./build.sh
```

failure:

```
... build output ...
[02:25:15] Begin /scratch/pi-gen/pi-gen/export-image
[02:25:15] Begin /scratch/pi-gen/pi-gen/export-image/prerun.sh
Mounting BOOT_DEV...
Mounting ROOT_DEV...
/boot: offset 4194304, length 268435456
/:     offset 272629760, length 1736441856
mkdosfs: Partitions or virtual mappings on device '/dev/loop0', not making filesystem (use -I to override)
```
```
debian@builder:/scratch/pi-gen/pi-gen$ sudo losetup -a
/dev/loop1: [0077]:194119 (/scratch/pi-gen/pi-gen/work/rpi-test-lite/export-image/2022-10-20-rpi-test-lite-lite.img), offset 272629760, sizelimit 1736441856
/dev/loop0: [0077]:194119 (/scratch/pi-gen/pi-gen/work/rpi-test-lite/export-image/2022-10-20-rpi-test-lite-lite.img), offset 4194304, sizelimit 268435456
debian@builder:/scratch/pi-gen/pi-gen$ grep loop /proc/partitions 
   7        0     262144 loop0
 259        0     262144 loop0p1
 259        1    1695744 loop0p2
   7        1    1695744 loop1
[...]

PR #641, opened 2022-10-22
use a single partitioned loop device

rfc6919:
per https://github.com/RPi-Distro/pi-gen/issues/639, use a single partitioned loop device rather than multiple devices with offset/length.

testing is limited to "works for me, and doesn't leave any new errors in the output of build.sh". is there an actual test suite or anything like that?


XECDesign:
I've merged this into the internal nightly build system. If it doesn't cause any issues, I'll merge the PR in a few days.

Thanks!

XECDesign:
Pushed a couple of fixes.

`unmount_image` returns false and halts the whole build if the image isn't mounted in the first place. Adding the `if` statement back rather than using `&&` should fix that.

Then if the loopback device is attached but the either partition is not mounted, you get `umount: /dev/loopXpN: not mounted.` and again, false is returned.

If the mounted partitions have subdirectories mounted, then `umount` isn't enough. Maybe an a recursive `umount` would be fine, but there's already an `unmount` option which should do the right thing.

I've also changed it to iterate over all the partitions rather than assuming there are only two.

XECDesign:
Seems to work now, thanks.

Issue #557, opened 2021-11-02
Image does not auto-resize during first boot when using qcow2?


jvonau:
Spun up an image using the default USE_QCOW2=1, while much faster and less messy during the build process the resulting image when booted failed to auto resize the root partition. Now setting USE_QCOW2=0 the resulting image resizes correctly during first boot. Anybody else seeing this behavior?

XECDesign:
The QCOW2 feature was added without proper review and seems to have caused more issues than it's worth. I'll accept PRs to fix it, but at this rate it looks like it's more likely to be removed entirely.

PR #648, opened 2022-11-13
Remove QCOW2 build mechanism

mmmspatz:
IMO bf8c9f5 added an undue amount of complexity, and [this](https://github.com/RPi-Distro/pi-gen/issues/557#issuecomment-1062787650) comment suggests that maintainers are open to getting rid of it. So I went ahead and did that.

Carefully revert bf8c9f5, making a few exceptions:

1. Keep some inconsequential formatting changes & line reordering. No reason to change it back.
2. Keep the change that mounts host `/dev` in the container during the docker build. This happens to have been introduced in bf8c9f5, but is necessary for loop devices to show up after `losetup` runs. See: https://github.com/moby/moby/issues/27886

I suggest viewing this commit with "hide whitespace changes" enabled.


XECDesign:
Many thanks for doing this.

I'll leave it for a bit to go over it and give people a chance to raise objections.

In the meantime, I'll implement a simpler alternative approach.

mmmspatz:
I took a stab at doing a similar thing in a much simpler way with btrfs subvolumes. It is just an optimization of `copy_previous()`; the final export still rsyncs files into filesystems backed by a loop device. It happens automatically if the host filesystem happens to be btrfs.

Here is the main commit: https://github.com/mmmspatz/pi-gen/commit/862219c8066c4fc51a93898ec06af7be5db074de
And [here](https://github.com/mmmspatz/pi-gen/tree/mspatz/copy_previous_btrfs) that is on a branch.
 
If this is of interest or appeases anyone, I can make a new PR for it. The speedup is pretty much nil for me (copying a handful of GB a handful of times does not take that long relative to everything else), but the disk usage savings are there. And maybe someone has a situation where the speedup would be noticeable.

An even simpler change than what I've done here would be to replace rsync with cp in `copy_previous()`, then we could pass `--reflink=auto` and get almost exactly the same benefit. rsync does not seem to have a similar feature.

mmmspatz:
Crossed my mind to bump this, so here I am doing so. Just rebased onto current master.

jakicoll:
Now that the qcow2 mechanism is broken on bookworm (#710), I guess this should really be merged in.

mmmspatz:
Just rebased onto current master again. Merge conflicts were straightforward to solve.

mmmspatz:
Another straightforward rebase onto master.

XECDesign:
Thanks and sorry about the delay. I just wanted to make sure there has been enough warning that this is happening.

PR #649, opened 2022-11-14
Remove resize2fs_once init script

mmmspatz:
As of 7206ab1, the [firstboot](https://github.com/RPi-Distro/raspberrypi-sys-mods/blob/master/usr/lib/raspberrypi-sys-mods/firstboot) script runs as init during first boot. Unlike init_resize.sh before it, firstboot runs `resize2fs` immediately after resizing the root partition. The `resize2fs_once` init script is therefore redundant. Shall we get rid of it?


sithmein:
Side note: issue #751 and the corresponding PR will add support for btrfs. btrfs can only be resized when the file system is mounted. Therefore a resize script outside the initrd is still useful.

sithmein:
https://github.com/RPi-Distro/pi-gen/pull/753 also removes this script.

mmmspatz:
Done by a078afeba001f93cc1c74a266bdd07d3f03ec713

PR #679, opened 2023-03-22
Check that docker is not set up in rootless mode

stefanb2:
By default I use Docker in rootless mode, i.e. the Docker daemon is not running as `root`. I have not been able to execute the docker build, because mounting of `binfmt_misc` fails:

~~~log
+ docker run --name pigen_work --privileged --cap-add=ALL ... pi-gen bash -e -o pipefail -c 'dpkg-reconfigure qemu-user-static &&
        # binfmt_misc is sometimes not mounted with debian bullseye image
        (mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc || true) &&
        cd /pi-gen; ./build.sh  &&
        rsync -av work/*/build.log deploy/'
mount: /proc/sys/fs/binfmt_misc: permission denied.
Module binfmt_misc not loaded in host
Please run:
  sudo modprobe binfmt_misc
~~~

The work around is to use `sudo docker`, but `build-docker.sh` doesn't allow me to specify that.

This PR changes two things in the script:

* it allows the user to override the docker command with the `DOCKER` environment variable
* if `${DOCKER} info` indicates rootless mode then fall back to `sudo docker`


XECDesign:
I don't use docker myself, so I'm happy to take your word for it. Okay to merge now?

PR #681, opened 2023-03-24
Remove binfmt misc deps from Docker build

stefanb2:
In Docker build `binfmt_misc` support is provided by the host, hence we do not need to install any of its dependencies inside the image.

DRY removal: now there is only on `docker run` command line to maintain in the script.

### DIFF
--- a/export-image/00-allow-rerun/00-run.sh
+++ b/export-image/00-allow-rerun/00-run.sh
@@ -1,7 +1,9 @@
 #!/bin/bash -e
 
-if [ ! -x "${ROOTFS_DIR}/usr/bin/qemu-arm-static" ]; then
-	cp /usr/bin/qemu-arm-static "${ROOTFS_DIR}/usr/bin/"
+if [ "${PIGEN_RUNNING_INSIDE_DOCKER}" != "1" ]; then
+  if [ ! -x "${ROOTFS_DIR}/usr/bin/qemu-arm-static" ]; then
+    cp /usr/bin/qemu-arm-static "${ROOTFS_DIR}/usr/bin/"
+  fi
 fi
 
 if [ -e "${ROOTFS_DIR}/etc/ld.so.preload" ]; then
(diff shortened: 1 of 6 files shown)


export-image/00-allow-rerun/00-run.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash -e
 
-if [ ! -x "${ROOTFS_DIR}/usr/bin/qemu-arm-static" ]; then
-	cp /usr/bin/qemu-arm-static "${ROOTFS_DIR}/usr/bin/"
+if [ "${PIGEN_RUNNING_INSIDE_DOCKER}" != "1" ]; then
+  if [ ! -x "${ROOTFS_DIR}/usr/bin/qemu-arm-static" ]; then
+    cp /usr/bin/qemu-arm-static "${ROOTFS_DIR}/usr/bin/"
```
stefanb2, line 5: I started to wonder about this step

* is that actually the correct command for a build on `arm64` branch? I.e. shouldn't that be `/usr/bin/qemu-aarch64-static` instead?
* shouldn't that step be skipped when running the build on a native platform?

stefanb2: On second thought: why is this `cp` required at all? `binfmt_misc` is executed by the kernel, so it would use the QEMU binary from the host root, not the copy in the chroot environment.

XECDesign: You're right on all points. In the past, the guest had to have qemu-arm-static available, but most distros now register qemu with binfmt_misc with the `F` flag. When that's used, the binary is pre-loaded in advance from the 'host'. If you wanted to be careful, you could check whether that's the case by reading /proc/sys/fs/binfmt_misc/qemu-{aarch64,arm}.

stefanb2: Well I'm happy with the flagging as it is. Maybe at some point that `cp` could simply be removed as being unnecessary.

stefanb2:
No matter what I try, it seems `docker build` doesn't want to import a build argument value into an environment variable: the variable always ends up with an empty string. Maybe I'm just too blind to see a dumb typo in my code.

**Shrug** I guess then we just have to live with the unnecessary dependencies in the Docker image.

PR #685, opened 2023-03-28
Register binfmt_misc entry for docker build

moh53n:
In some distros (like some archlinux based distros) qemu-user-static package doesn't register qemu for binfmt and causes problem. [multiarch/qemu-user-static](https://github.com/multiarch/qemu-user-static) solves this by a simple privileged docker run. AFAIK it doesn't make any conflict with existing qemu-user-static packages and it's a universal solution. I think it's good to use this in the docker build script in order to prevent possible problems on various distros.

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -107,6 +107,40 @@ else
   )
 fi
 
+# Check if binfmt_misc is required
+binfmt_misc_required=1
+case $(uname -m) in
+  aarch64)
+    binfmt_misc_required=0
+    ;;
+  arm*)
+    binfmt_misc_required=0
+    ;;
+esac
+
+# Check if qemu-arm-static and /proc/sys/fs/binfmt_misc are present
+if [[ "${binfmt_misc_required}" == "1" ]]; then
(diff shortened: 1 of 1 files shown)


moh53n:
I agree. There is also another approach that makes sure binfmt works without relying on a third party; I'm going to update the pull request.

moh53n:
Now, we check if qemu-arm-static and /proc/sys/fs/binfmt_misc are present and then register a binfmt_misc entry to ensure qemu-arm-static is registered (binfmt_misc won't care duplicate entries unless they have common names).

XECDesign:
Cheers

JavanXD:
This PR broke my build-docker.sh command which I've been using for years on MacOS. The Shell script exits with the error message `qemu-arm-static not found (please install qemu-user-static)` although the build is working fine the versions before the PR. I am using MacOS Sonoma Version 14.0 (23A344). I have not found any way to install the missing package on MacOS so I moved back in time using a few tags before `git checkout 2023-02-21-raspios-bullseye`.

PR #686, opened 2023-03-28
DRY removal for docker run command line

stefanb2:
Factor out the differences into variables so that there is only one implementation of the `docker run` command line to maintain.


XECDesign:
Thanks

PR #693, opened 2023-05-09
Skip binfmt_misc register when unnecessary

stefanb2:
This improves commit 6dc45a80e764eacd6e311c32164b623bbe7f8e08

Check if interpreter is already registered before running the register step. This avoids unnecessary "sudo" execution which may require user interaction for the password.

Improves on #685


stefanb2:
@moh53n can you please check this as you are the author of the original PR?

stefanb2:
No idea what the `|| true` is supposed to achieve, but restored it...

stefanb2:
**FYI** I also have a PR which adapts #685 for the `arm64` branch, but GitHub is having issues right now...

moh53n:
LGTM, Also, the `|| true` part eliminates error when our interpreter is already registered (with the same name).

XECDesign:
Thanks

PR #694, opened 2023-05-09
Register aarch64 in binfmt misc

stefanb2:
This modifies commit 6dc45a80e764eacd6e311c32164b623bbe7f8e08 for the
arm64 branch. For 64-bit builds we need to register qemu-aarch64-static
instead.

Improves on #685


stefanb2:
**NOTE** this change *only* applies to branch `arm64` and should not be merged to `master`!

moh53n:
LGTM.

PR #700, opened 2023-07-24
add backing image format as required by newer qemu

asssaf:
qemu-img requires specifying the backing image format (since [6.1](https://wiki.qemu.org/ChangeLog/6.1))


XECDesign:
I'm not sure that it's worth making the change now, since qcow based builds are going to be removed,

PR #705, opened 2023-09-05
fix: support bash v3 by removing array usage

reubenmiller:
Improve compatibility of the `docker-compose.yaml` script by supporting bash v3. MacOS users only have bash v3 by default (due to licensing).

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -1,4 +1,6 @@
-#!/bin/bash -eu
+#!/usr/bin/env bash
+# Note: Avoid usage of arrays as MacOS users have an older version of bash (v3.x) which does not supports arrays
+set -eu
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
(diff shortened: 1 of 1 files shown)


build-docker.sh
```diff
@@ -1,4 +1,6 @@
-#!/bin/bash -eu
+#!/usr/bin/env bash
+# Note: Avoid usage of arrays as MacOS users have an older version of bash (v3.x) which does not supports arrays
+set -eu
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
XECDesign, line 5: The use of `${BASH_SOURCE[0]}` isn't an issue?

reubenmiller: Yes you're right, good catch.

reubenmiller: Resolved by 6b630b400223fcaa24101e681b8716ba51df069e

XECDesign:
I'll leave this hanging for a little bit, since I'm not a Docker or MacOS user. Hoping that will give time for people who are to provide some feedback.

rhysackerman:
> I'll leave this hanging for a little bit, since I'm not a Docker or MacOS user. Hoping that will give time for people who are to provide some feedback.

Tested and it seemed to work fine for me. macOS Ventura, latest Docker version. Image generated fine, however failed on export due to running out of storage (my fault)

cjcdev:
Did this commit break using "PRESERVE_CONTAINER=1 CONTINUE=1" ?

PR #716, opened 2023-10-17
Fix build-docker.sh CONTINUE=1 mode

stefanb2:
This fix in commit 77706bd2f6b77fbdbc329267f8386a38e4704c34 broke the `CONTINUE=1` mode in `docker-build.sh`

Without the fix the Docker looks for the container `"pigen_work"`, not `pigen_work`.

**NOTE:** this PR needs to be merged to `master` & `arm64` as the fix was applied to both.


XECDesign:
Thanks!

PR #729, opened 2023-11-26
Enable mouse_snap in Wayfire

probonopd:
This PR enables snapping windows then moved to screen borders. This is a feature users have come to expect from other operating systems and desktop environments.

Reference:
* https://github.com/WayfireWM/wayfire/discussions/2018


XECDesign:
pi-gen isn't meant to be overriding that kind of thing.

The PR should be against this file, I think: https://github.com/raspberrypi-ui/raspberrypi-ui-mods/blob/master/etc/wayfire/defaults.ini#L21

But, I'm guessing the current behaviour is what's intended (@spl237?).

PR #741, opened 2024-01-03
Ensure loop device partition nodes are created

### LINKED ISSUE #482: losetup fails with docker-in-docker

acinonyx:
Although the loop block device is created before attaching the image to it, the devices for the partition that the image contains are still not created. This patch creates those devices as well, when they are not already available.

Fixes #482

### DIFF
--- a/export-image/prerun.sh
+++ b/export-image/prerun.sh
@@ -46,6 +46,7 @@ if [ "${NO_PRERUN_QCOW2}" = "0" ]; then
 		fi
 	done
 
+	ensure_loopdev_partitions "$LOOP_DEV"
 	BOOT_DEV="${LOOP_DEV}p1"
 	ROOT_DEV="${LOOP_DEV}p2"
 
--- a/export-noobs/prerun.sh
+++ b/export-noobs/prerun.sh
@@ -22,6 +22,7 @@ until ensure_next_loopdev && LOOP_DEV="$(losetup --show --find --partscan "$IMG_
 	fi
 done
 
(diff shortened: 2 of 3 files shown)


bamartin125:
@acinonyx 

I have not investigated why this is needed, but in checking out your PR, I noticed that it only proposes a functional change to `export-image/prerun.sh`. Is a change not needed to `export-noobs/prerun.sh` like #483 introduced? I just don't know offhand and I don't intend to investigate now.

acinonyx:
Thanks for taking the time to take a look on the PR. Indeed, that was an oversight on my part. I've updated the commit.

I don't know if it's useful but [here](https://cdn.artifacts.gitlab-static.net/60/43/6043d70cbd45a69a3752cdb5bce1d443cf4a8e8ec76a1590e12e5da39422e33e/2024_01_01/5841237772/6341478965/job.log?response-content-type=text%2Fplain%3B%20charset%3Dutf-8&response-content-disposition=inline&Expires=1704296713&KeyName=gprd-artifacts-cdn&Signature=pTlNpvBeK6uekbkDNsGvnDYAIUk=) you can find logs of a failed run. The commit of this pipeline is [this](https://gitlab.com/acinonyx/satnogs-pi-gen/-/commit/01c4dff4c3f383e4dcd1cd794bd36429fa2a8b33). The GitLab runner pipeline runs with a Docker-in-Docker setup.

XECDesign:
Thanks!

PR #748, opened 2024-01-24
Ensure loop device partition nodes are created (arm64 branch) (#741)

### LINKED ISSUE #482: losetup fails with docker-in-docker

jeffmakes:
Apply @acinonyx 's fix to the arm64 branch. I've tested this on an arm64 AWS instance where I'm developing a custom build. Apologies if I'm doing this wrong - this is only my ~third PR!

Although the loop block device is created before attaching the image to it, the devices for the partition that the image contains are still not created. This patch creates those devices as well, when they are not already available.

Fixes #482

### DIFF
--- a/export-image/prerun.sh
+++ b/export-image/prerun.sh
@@ -46,6 +46,7 @@ if [ "${NO_PRERUN_QCOW2}" = "0" ]; then
 		fi
 	done
 
+	ensure_loopdev_partitions "$LOOP_DEV"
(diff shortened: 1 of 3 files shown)


XECDesign:
Sorry, I should've merged master to arm64 and pushed the changes earlier. Is the version in arm64 working for you now?

jeffmakes:
Yep, I've just tested it and it builds ok.

Issue #751, opened 2024-02-10
Use gpt and btrfs as root filesystem


sithmein:
It should be possible to use btrfs instead of ext4 as the root file system. There are several instruction on the internet how an existing installation can be converted but they are complicated and error-prone. Therefore having an image that uses btrfs in the first place would be quite useful.
In addition using storage devices with more then 2TB (the Raspberry Pi 5's new PCIe interface makes this possible easily) is currently not possible since MBR does not support partitions with more than 2TB. Therefore changing to gpt is required.

XECDesign:
GPT is not officially supported by the bootloader and Raspberry Pi OS currently only supports ext4 or nfs.

We may come back to btrfs at some point in the future, but that's not currently on the task list.

sithmein:
What exactly do you mean with _only supports ext4 or nfs_? btrfs clearly works on Raspberry Pi OS.

PR #753, opened 2024-02-11
Support for btrfs and gpt

sithmein:
This PR adds two new configuration options - `FILE_SYSTEM_TYPE` and `PARTITION_TABLE_TYPE` - which allow building images with btrfs (instead of ext4) and/or gpt (instead of msdos/MBR).

This fixed issue #751.


XECDesign:
See https://github.com/RPi-Distro/pi-gen/issues/751#issuecomment-1941554706

PR #765, opened 2024-04-25
export stage list

knachte:
This makes stage_list available in the individual stages, which enables adding packages depending on which stages are included.


XECDesign:
Thanks

Issue #770, opened 2024-05-01
If source directory has been bind-mounted, build breaks


pevsonic:
On my Debian 12 workstation, the main large drive `/bigssd` has various subdirectories (e.g. `projects`, `home`, `opt` ...) that get bind mounted from the disks mount to their actual mount points on the filesystem. This has an interesting side-effect that when pi-gen mounts subdirs under $ROOTFS_DIR, mount reports them twice, for example :

```
proc on /workdrive/projects/pi-gen/work/raspios/export-image/rootfs/proc type proc (rw,relatime)
proc on /mnt/bigssd/workdrive/projects/pi-gen/work/raspios/export-image/rootfs/proc type proc (rw,relatime)
```

This would be OK, except that when you umount one, both become unmounted. In `scripts/common:unmount()` this will cause an error.

I think that potentially calling `umount -f` might bypass this, but I think it would be better to construct the lists of directories to unmount more carefully instead.

pevsonic:
Fixed this with a pleasingly minimal change here at : https://github.com/pevsonic/pi-gen/commit/2fe8d477436940abaee4f010a7e0df1b3654cd75

XECDesign:
Thanks!

So I understand a bit better, could you please give an example output lines of 'mount' which demonstrate the situation where the problem would occur?

pevsonic:
> So I understand a bit better, could you please give an example output lines of 'mount' which demonstrate the situation where the problem would occur?

Sure!

So test scenario... on this machine, physical drive (bigssd) is mounted at `/mnt/bigssd`. It has a top level split of directories that get bind-mounted to their actual locations :

```
/home
/opt
/tftpboot
/workdrive
...
```

For example, fstab contains :
```
#
# Bind-mounts
#

/mnt/bigssd/home					/home		auto	defaults,bind	0	0
/mnt/bigssd/opt						/opt		auto	defaults,bind	0	0
/mnt/bigssd/workdrive					/workdrive	auto	defaults,bind	0	0
/mnt/bigssd/tftpboot					/tftpboot	auto	defaults,bind	0	0
```

I create a new test project under `/workdrive/projects` which is part of my usual workflow :

```
cd /workdrive/projects
git clone  -b arm64 https://github.com/RPi-Distro/pi-gen.git
cd pi-gen
```

At the time of writing, `git log --oneline  | head -1` gives me : 
```
64e2057 Merge branch 'master' into arm64
```

[...]

PR #771, opened 2024-05-01
Compare mount output using base of dirname

pevsonic:
If you're building from a bind-mounted directory, the build will fail as it will find two entries to unmount, but a single unmount will remove them both causing an error. Adding a space means that the mountpoint will only match with a single mount entry ; the expected path, rather than the pre bind-mount.

Fixes issue #770

### DIFF
--- a/export-image/05-finalise/01-run.sh
+++ b/export-image/05-finalise/01-run.sh
@@ -83,7 +83,7 @@ cp "$ROOTFS_DIR/etc/rpi-issue" "$INFO_FILE"
 	dpkg -l --root "$ROOTFS_DIR"
 } >> "$INFO_FILE"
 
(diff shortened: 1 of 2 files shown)


XECDesign:
Many thanks! I've reproduced the issue you were seeing and this fixes it for me. I've made some additional changes to the PR while I was in the area. Could you please give this version a go and see if it still works for you as expected?

Normally changes go to the master branch and are then merged into arm64, so I've changed the base of the PR. You might need to merge it back into your arm64-based branch to test it.

Quite often people will try to rm -rf the work directory while there are some mountpoints there, which can cause all kinds of issues on the system, so I've added a warning not to do that.

Combined the previous implementation of `unmount` into a single while loop.

Added 5 retries so that if something goes wrong, it's not stuck in a loop trying to unmount forever.

PR #773, opened 2024-05-02
Expand trap cleanup function

tdewey-rpi:
Add a function to unmount the bindmounts in the chroot, and unmount the image and stage in progress, in the event of a trap.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -158,7 +158,12 @@ do
 done
 
 term() {
-	true; #TODO: Cleanup
+	if [ -n "${IMG_FILE}" ]; then
+		unmount_image "${IMG_FILE}"
+	fi
+	unmount "${WORK_DIR}/${STAGE}"
+	off_chroot
+	true;
 }
 
 trap term EXIT INT TERM
--- a/build.sh
+++ b/build.sh
@@ -158,12 +158,17 @@ do
 done
 
 term() {
-	if [ -n "${IMG_FILE}" ]; then
-		unmount_image "${IMG_FILE}"
+	if [ "$?" -ne 0 ]; then
+		log "Build failed"
+	else
+		log "Build finished"
 	fi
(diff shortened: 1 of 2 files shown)


tdewey-rpi:
Wait, what? I can't see why this has brought in those other two commits....

tdewey-rpi:
Paging @XECDesign for a view.

tdewey-rpi:
Confirmed I see this more gracefully handling fatal errors

XECDesign:
Is `off_chroot` needed? 

I believe the `unmount "${WORK_DIR}/${STAGE}"` line should already do everything the `off_chroot` function would've done.

The `on` in `on_chroot` maybe should've been `in`. The idea is that you pipe commands into it and it runs in the chroot environment.

Or have I missed something?

tdewey-rpi:
> Is `off_chroot` needed?
> 
> I believe the `unmount "${WORK_DIR}/${STAGE}"` line should already do everything the `off_chroot` function would've done.
> 
> The `on` in `on_chroot` maybe should've been `in`. The idea is that you pipe commands into it and it runs in the chroot environment.
> 
> Or have I missed something?

I found without this, I was still getting sub-mounts mounted on abort - which then caused trouble when I tried delete the workdir and re-try my build.

RE: naming:

Completely my fault for misreading. Happy to rename.

XECDesign:
This version, which also includes #771, works for me without `off_chroot` with failures introduced in various sub-stages and also in export-image while the img file is attached.

Rebased off the `master` branch, since that's what gets merge into `arm64`

The current git history is just for testing and clarity, so I've converted it to a draft for now

XECDesign:
Also, thanks! This has been a long overdue addition.

tdewey-rpi:
@XECDesign Yours to merge, I lack the powers for now.

Issue #785, opened 2024-07-28
Improvement:  Build Default Image using a GitHub Action


theGOTOguy:
For my own custom build utilizing Pi-Gen, I have written [a GitHub action](https://github.com/simpleairquality/SimpleAQ-Device/blob/main/.github/workflows/build.yml) that automatically builds the image, which is then available to download directly through GitHub.

This would:

1. Automatically confirm on each push that the build process isn't broken.
2. Provide users with a ready-built Raspbian image with the most common settings for download and testing.
3. Serve as a blueprint for others who might like to build custom images based on Raspbian.

Please feel free to use my existing, open-source action above at a starting point.  Or, if there is interest from the devs, I will be happy to contribute a PR for you that would build a Raspbian image with the most common settings.

XECDesign:
That's great, thanks!

We use Gitlab CI for our official builds, but having a github workflow as an example for others would be useful.

I'd merge a PR that builds the default images without customisation.

I'm guessing Github doesn't provide enough disk space to build non-lite images, but that problem should be solved when I add COW support for stage copies.

jedahan:
Here's [another example workflow](https://github.com/alexnathanson/solar-protocol/blob/main/.github/workflows/main.yml) and the [build-image.sh](https://github.com/alexnathanson/solar-protocol/blob/main/utilities/build/build-image.sh) that runs inside the build step. Couldn't have gotten this far without @theGOTOguy work, so wanted to reciprocate.

SvenFinn:
Just to add my two cents:

I've run pi-gen builds on Github Actions for a project before, running into - seemingly - the same limitations as you regarding github actions storage.
During my project we modified the copy_previous function to delete the work directory of the previous build step after copying it, to reduce the amount  of (unnecessary) space taken up by older stages.

This has two major drawbacks:

- The EXPORT_IMAGE step is ran at the end of the build, using the work directories of each step that is to be exported. Deleting the work directory of a step makes it impossible to export this step. Therefore, when using this space optimization, you can only export the last build stage to an image.
- If the build fails, you cannot reuse the state, as the work directories of all previous stages are missing. When building using GitHub Actions, I don't think this is an issue.

To avoid the first drawback, you could restructure pi-gen to export images after it finishes building the stage, but that would require restructuring pi-gen.

PR #786, opened 2024-07-31
Adds GitHub Action to Build Raspbian Lite

theGOTOguy:
This action uses entirely default settings, and disables stage3-stage5 builds.  I was unable to get any of these builds to complete within the space constraints of the GitHub runners, but if they add disk space to their runners it may be possible to build these in the future.

Tested and confirmed working on a Raspberry Pi Zero W.

Completes #785

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,7 +52,7 @@ jobs:
 
     - name: Build Image
       run: |
-        ls
+        chmod a+x build-docker.sh 
         ./build-docker.sh
 
     - name: Get Image Name
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Maximize Build Space
       uses: easimon/maximize-build-space@master
       with:
-        root-reserve-mb: 25000 
+        root-reserve-mb: 20000 
         swap-size-mb: 1024
         remove-dotnet: 'true'
         remove-android: 'true'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Maximize Build Space
       uses: easimon/maximize-build-space@master
       with:
-        root-reserve-mb: 10240
+        root-reserve-mb: 4096 
         swap-size-mb: 1024
         remove-dotnet: 'true'
         remove-android: 'true'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,6 +38,7 @@ jobs:
     - name: Create Config File
       id: config
       run: |
+        touch config
         NOW=$(date +"%Y-%m-%d-%H%M")
         IMAGE=Raspbian-${GITHUB_REF##*/}-$NOW
         echo "image=$IMAGE"                        >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+        remove-haskell: 'true'
+        remove-codeql: 'true'
+        remove-docker-images: 'true'
+
+    - name: List Available Space
+      run: |
+        echo "Free space:"
+        df -h
+
+    - name: Install Dependencies
```
usimd, line 25: You could reuse the [declared build dependencies](https://github.com/RPi-Distro/pi-gen/blob/master/depends) instead of listing them manually.

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+      run: |
+        sudo apt update
+        sudo apt install -y coreutils quilt parted qemu-user-static \
+                            debootstrap zerofree zip dosfstools     \
+                            libarchive-tools libcap2-bin grep rsync \
+                            xz-utils file git curl bc qemu-utils    \
+                            kpartx gpg
+
+    - name: Checkout PiGen
+      uses: actions/checkout@v3
```
usimd, line 35: `actions/checkout` is `v4` already.

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+      id: imagefile
+      run: |
+        cd deploy
+        ls
+        IMAGE_FILE=$(ls *.zip)
+        echo "imagefile=$IMAGE_FILE" >> $GITHUB_OUTPUT
+
+    # The image now exists in deploy/. 
+    # Make the artifact available from the action.
+    - uses: actions/upload-artifact@v1
```
usimd, line 68: `actions/upload-artifact` is also at `v4`, this version is heavily outdated.

usimd (commented):
Cool idea to increase the disk space for building. Thinking about adding this to [`pi-gen-action`](https://github.com/usimd/pi-gen-action) as well.

.github/workflows/build.yml
theGOTOguy: Updated to `v4`.

.github/workflows/build.yml
theGOTOguy: Good call.  Mostly using `depends` now, but there was one additional needed dependency, `qemu-user-static`.  Not sure whether the right thing is to put it into `depends`, but for now it's installed separately as part of the build step.

.github/workflows/build.yml
theGOTOguy: Updated to `v4`.

theGOTOguy:
Updated dependency install and `upload-artifact` and `checkout` versions.

Build confirmed working:  https://github.com/theGOTOguy/pi-gen/actions/runs/10175875697

.github/workflows/build.yml
usimd: Maybe @XECDesign can elaborate on whether this could be added to `depends` as well.

usimd:
FWIW, I've released a new version of [`pi-gen-action`](https://github.com/usimd/pi-gen-action) that includes a similar but more aggressive mechanism as in [`easimon/maximize-build-space`](https://github.com/easimon/maximize-build-space). 

With this option turned on (`increase-runner-disk-size: true`) I was able to build a full image up to stage5 on a free GH runner. If you're curious and maybe want to extend your change, have a look at the additional resources to free: 

https://github.com/usimd/pi-gen-action/blob/afb6cdd932ef6e17cc3850ace040df9430325065/src/increase-runner-disk-size.ts#L13-L102

PR #790, opened 2024-09-16
Additions to assist with pi-gen image customisation

learmj:
Hi @XECDesign 
Here's a small set of commits plus introduction of an examples directory to help jump start projects, or to highlight anything in pi-gen we think would be valuable to users. We can add to examples/ as we see fit.

Supporting a postrun hook when processing stages allows a user to execute operations in that stage after all sub-stage actions have been executed.

The addition of EXPORT_CONFIG_DIR allows a user to define their own actions for image creation rather than using/editing the defaults. By using STAGE_LIST and EXPORT_CONFIG_DIR, a user has complete control over their image construction and generation without having to worry about rebasing/merging things when we make changes in the defaults.

The addition of bootstrap64 is to add a 64-bit debootstrap install wrapper so that a user can create a 64-bit custom image without having to switch to the arm64 branch. I think whoever invokes this (eg from a custom stage) should take responsibility of being able to run it on the Host, or I can a 'suitability test' to it if you'd prefer.

examples/min64 is a generic 64-bit target system with limited functionality, intended to serve as an example for users wishing to create a light-weight system.

Comments and feedback welcome.
Cheers,
--  Matt


timg236:
@XECDesign @tdewey-rpi - Please could you take a look at this ?

examples/min64/export-image/04-finalise/01-run.sh
```diff
@@ -0,0 +1,118 @@
...
+
+rm -f "${ROOTFS_DIR}/usr/bin/qemu-arm-static"
+
+if [ "${USE_QEMU}" != "1" ]; then
+	if [ -e "${ROOTFS_DIR}/etc/ld.so.preload.disabled" ]; then
+		mv "${ROOTFS_DIR}/etc/ld.so.preload.disabled" "${ROOTFS_DIR}/etc/ld.so.preload"
+	fi
+fi
+
+rm -f "${ROOTFS_DIR}/etc/network/interfaces.dpkg-old"
```
tdewey-rpi, line 30: I'm not super-keen on this mechanism, because you wind up with 'export image' performing intrusive modifications on the system that may produce effects that you'll only see in later QA.

I'd suggest exploring the mechanism we deployed in pi-gen-micro: https://github.com/raspberrypi/pi-gen-micro/blob/main/configurations/cryptroot_image/dpkg_extra_args

Essentially, that mechanism prevents files matching a globbing pattern you specify from _ever being installed_. This is handy, because it means you can also have correctness checks as part of your regular build flow. Performing the modifications here doesn't really allow for that without re-mounting the image, or running a second step after pi-gen.

tdewey-rpi:
Noble stab at improvements, and certainly something I think we could accept now to build on.

I do have one fairly-large question, though: What if we just made EXPORT_IMAGE a regular stage? So rather than using something like...

`STAGES="stage1 stage2 stage3"`

You could instead explicitly label as part of your command set where you want images to be taken:

`STAGES="stage1 export_image stage2 export_image stage3 export_image"`

examples/min64/export-image/04-finalise/01-run.sh
learmj: > I'm not super-keen on this mechanism, because you wind up with 'export image' performing intrusive modifications on the system that may produce effects that you'll only see in later QA.
> 
> I'd suggest exploring the mechanism we deployed in pi-gen-micro: https://github.com/raspberrypi/pi-gen-micro/blob/main/configurations/cryptroot_image/dpkg_extra_args

I just copied what we currently do today, but I agree.

learmj:
> Noble stab at improvements, and certainly something I think we could accept now to build on.
> 
> I do have one fairly-large question, though: What if we just made EXPORT_IMAGE a regular stage? So rather than using something like...
> 
> `STAGES="stage1 stage2 stage3"`
> 
> You could instead explicitly label as part of your command set where you want images to be taken:
> 
> `STAGES="stage1 export_image stage2 export_image stage3 export_image"`

By using new variable EXPORT_CONFIG_DIR, you can specify any directory to hold the export operations. I believe STAGE_LIST is only used for generation rather for generation or export.
BTW I have a new example which is just an export. I'll add you to the PR.

tdewey-rpi:
> > Noble stab at improvements, and certainly something I think we could accept now to build on.
> > I do have one fairly-large question, though: What if we just made EXPORT_IMAGE a regular stage? So rather than using something like...
> > `STAGES="stage1 stage2 stage3"`
> > You could instead explicitly label as part of your command set where you want images to be taken:
> > `STAGES="stage1 export_image stage2 export_image stage3 export_image"`
> 
> By using new variable EXPORT_CONFIG_DIR, you can specify any directory to hold the export operations. I believe STAGE_LIST is only used for generation rather for generation or export. BTW I have a new example which is just an export. I'll add you to the PR.

Absolutely - but you still wind up with a two-part modification:

1. You must create an export operations directory
2. You must insert the magic export marker file into each stage as you require

The key issue is that you can't trivially compose systems with mixes of stages - because you need to modify the stage by adding that export marker.

If, however, you transform the export operations into a regular pi-gen stage, you're free to have one or more in any location you choose at invocation time - and all the usual mechanisms for version controlling a stage list still work as expected. You can chop & change your image by swapping out stages without any major concerns.

learmj:
> You could instead explicitly label as part of your command set where you want images to be taken:
> 
> `STAGES="stage1 export_image stage2 export_image stage3 export_image"`

That's reasonable, and perhaps more obvious from a flow perspective, but it would be at the expense of losing the flexibility that exists today by using ```<stage>/EXPORT_IMAGE``` because that file can add other attributes, eg set variables that are used for the export phase (eg IMG_SUFFIX as used today).

learmj:
Superseded by #798

PR #791, opened 2024-09-19
Dockerfile: fix ENV warning

stefanb2:
Newer versions of docker generate the following warning

   - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 4)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ARG BASE_IMAGE=debian:bullseye
 FROM ${BASE_IMAGE}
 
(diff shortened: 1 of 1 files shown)


cmorganBE (approved):
Made the same changes here due to the same warning. Would be good to get this PR upstream.

XECDesign:
And the new format still works fine with old versions of docker?

XECDesign:
> Or do you define "old version" as "we need to support all Docker releases since the initial release in 2013"?

Nope, just something that somebody might be justified running in 2025.

Thanks for checking.

PR #797, opened 2024-10-09
Cloud-Init support

paulober:
This PR adds basic cloud-init support to Raspberry Pi OS.


stage2/04-cloud-init/01-run.sh
```diff
@@ -0,0 +1,33 @@
...
+install -v -D -m 644 -t "${ROOTFS_DIR}/usr/lib/netplan/" files/00-network-manager-all.yaml
+
+# still does not solve the conflict, maybe some kind of race cond.
+# make sure config stage is run before userconfig service
+#sed -i '/^\[Unit\]/a Before=userconfig.service' "${ROOTFS_DIR}/lib/systemd/system/cloud-config.service"
+
+install -v -m 755 files/cloud-init-custom.deb "${ROOTFS_DIR}/tmp/cloud-init.deb"
+
+on_chroot << EOF
+	SUDO_USER="${FIRST_USER_NAME}" dpkg -i /tmp/cloud-init.deb || true
```
XECDesign, line 27: This won't be relevant when we update the package in our repo, so this is just a general heads up. Instead of using `dpkg -i` and following it up with `apt`, you could use `apt-get install -y /tmp/cloud-init.deb` directly and that will resolve dependencies and install the relevant packages in one step. The trick is that apt-get assumes that you're giving it package names unless it detects that you're giving it a path. I'm guessing it's checking for the presence of a `/` character, so if the file is in the current directory you'd use `./file.deb`.

You could probably even just put '/tmp/cloud-init.deb' in  `02-packages`, but I haven't tried that, so I'm not sure.

As is, using `|| true' means you ignore potential legitimate errors. Also, we normally nudge people towards `apt` rather than `apt-get` because it's more user friendly and handles some things automatically which `apt-get` doesn't, but in non-interactive scripts such as this one, `apt-get` is still the way to go. In other words, `apt-get` for scripting and `apt` otherwise.

paulober: Okay, thanks for explaining. I'll update this for the current state.

paulober: Just tried it and when using `apt-get install -y /tmp/cloud-init.deb` I got a message `Note: using cloud-init instead of /tmp/cloud-init.deb`

XECDesign: I'm guessing the package was updated without bumping the version number?

paulober:
Ref to required PR before merge: https://github.com/canonical/cloud-init/pull/5827

> Also requires custom NetworkManager and custom cloud-init package in Raspberry Pi OS repositories.

PR #802, opened 2024-10-21
Allow export image customisation

learmj:
Add new variable EXPORT_CONFIG_DIR to set the location of the scripts pigen will run when exporting an image. Setting this is optional. If not specified, the current location is retained.

By utilising STAGE_LIST AND EXPORT_CONFIG_DIR, a user can construct custom images out-of-tree without modification to any defaults.

(cherry picked from commit e5e6ceeaf46f52f77b759d3d35aef8bbd0a69c8b)


XECDesign:
We normally merge the master branch into arm64, so #801 is enough. Thanks!

PR #803, opened 2024-11-02
config.txt: update bad link and path to README

graysky2:
correct several typos in config.txt


graysky2:
> What is the problem with http://rptl.io/configtxt?
It did not work for me when I opened this PR, but it is working now.

> Also /boot/firmware/overlays/README is the correct URL on RPiOS.

Sorry, thought that path was the same on RPiOS as it is on Arch ARM.

PR #805, opened 2024-11-04
fix: Adjust execution order to resolve dns issue in user-rename step

Dhanus3133:
This PR resolves package installation failure during the `rename-user` step of the `export-image` sage. The error occurred due to a temporary DNS resolution issue when fetching the `userconf-pi` package. ([Error Actions URL](https://github.com/Dhanus3133/raspap-webgui/actions/runs/11668582128/job/32488575043#step:4:6328))

#### Changes Made

- **Reordered Scripts:** Network configuration tasks were prioritized before package installation to ensure proper DNS resolution.

**Successful Run After Changes:** [Actions URL](https://github.com/Dhanus3133/raspap-webgui/actions/runs/11668566995/job/32488526621#step:4:6512)


XECDesign:
Seems like something else is going on there. How could you get to the export stage without working DNS in the first place? I believe the initial resolv.conf file is copied from the host.

Dhanus3133:
> Seems like something else is going on there. How could you get to the export stage without working DNS in the first place? I believe the initial resolv.conf file is copied from the host.

```sh
curl -sL https://install.raspap.com | bash -s -- --yes --openvpn 1 --restapi 1 --adblock 1 --wireguard 1 --tcp-bbr 1
```
Before the export-image stage, the previous stages are executed without OpenVPN enabled, which means DNS relies on the host’s configuration. Since OpenVPN isn’t established until the [custom stage](https://github.com/Dhanus3133/raspap-webgui/blob/test/.github/workflows/release.yml#L31-L33), enabling it here leads to issues when exporting the image. 

In this commit ([7ed5a9f5](https://github.com/Dhanus3133/raspap-webgui/commit/7ed5a9f5dd4b988527a9b9a6206dff1ae8851d2a)), I’ve disabled OpenVPN flags and thus the export-image stage builds are successful.

XECDesign:
Yeah, if something else leaves things in a broken state, I don't think it's an ordering issue. It's probably best not to change the DNS settings until the very end, as it currently is.

PR #833, opened 2025-03-12
Align partition start boundaries to 8M

P33M:
Higher-capacity cards (particularly Raspberry Pi 128GB A2) have an 8MB preferred erase size, which is a good indication of the flash block size.

Other large cards may also benefit.


XECDesign:
Thanks!

PR #834, opened 2025-03-14
export-image: generate bmap file if bmaptool is available

roliver-rpi:


roliver-rpi:
Hi @XECDesign , this change generates bmaps of image files whilst they are still available in sparse form (i.e. before compression).

This change was tested on a build of the 'lite' image last night and resulted in a generated bmap which only mapped ~76.5% of blocks in the expanded image.

Preserving this sparse information will potentially allow for more optimised flashing of images or conversion to other sparse image formats if required.

XECDesign:
Great, thanks!

PR #863, opened 2025-08-24
Update archive keys from raspberrypi-archive-keyring package

### LINKED ISSUE #862: Repository copy of Archive Signing Key uses SHA-1 signature
The 'Raspberry Pi Archive Signing Key':
```
pub   rsa2048 2012-06-17 [SC]
      CF8A 1AF5 02A2 AA2D 763B  AE7E 82B1 2992 7FA3 303E
uid           [ unknown] Raspberry Pi Archive Signing Key
sub   rsa2048 2012-06-17 [E]
      DE5C 1FD2 8F18 8253 91E9  2B58 F7BB 0489 EDD8 3D6C
```

as included in `stage0/files/raspberrypi.gpg`  use SHA-1 signatures (exclusively).

This is probably not too problematic security-wise if you copied the proper key (i.e. the one with CF8A1AF502A2AA2D763BAE7E82B129927FA3303E fingerprint). The subkey is used only for encryption, which is not a function used by an Archive key.

Yet, it's enough to make debootstrap fail on a modern OS:

```
Begin /root/pi-gen/stage0/prerun.sh
I: Target architecture can be executed
I: Retrieving InRelease 
I: Checking Release signature
E: Invalid Release signature
```

debootstrap.log shows:
```
Signing key on A0DA38D0D76E8B5D638872819165938D90FDDD2E is not bound:
           No binding signature at time 2025-08-23T22:30:30Z
  because: Policy rejected non-revocation signature (PositiveCertification) requiring second pre-image resista
nce
  because: SHA1 is not considered secure since 2023-02-01T00:00:00Z
```

Steps to reproduce:
* Run pi-gen on a vanilla trixie image for armhf architecture
(oddly, it does validate it in the arm64)

<s>Steps to fix it:
1. Resign the key
```
gpg -u A0DA38D0D76E8B5D638872819165938D90FDDD2E --force-sign-key  --quick-sign-key A0DA38D0D76E8B5D638872819165938D90FDDD2E
```
the `--force-sign-key` parameter is needed if you have gpg < 2.5.7 (see see https://dev.gnupg.org/T7663)

2. Export the updated key and publish it
```
gpg -o raspberrypi.gpg --export A0DA38D0D76E8B5D638872819165938D90FDDD2E
gpg --armor -o raspberrypi.gpg.key --export A0DA38D0D76E8B5D638872819165938D90FDDD2E
```

Ideally, the final trixie images would ship already with the updated key.</s>

The `raspberrypi-archive-keyring` already has the updated keys, so it should be just a matter of updating `stage0/files/raspberrypi.gpg` on `pi-gen` repository (and, while we are at it, http://archive.raspberrypi.com/debian/raspberrypi.gpg.key which is the same version that was in the repo)

I have confirmed that updating `stage0/files/raspberrypi.gpg from `raspberrypi-archive-keyring` solves the issue.

PS: It is also possible to workaround this by changing the crypto policy to always trust sha1, but the real fix is to update the key signature, which is harmless (in comparison, an upgrade to a bigger rsa key would be much more complicated).

Keisial:
Use keys with SHA256 signatures

Fixes #862


florian-asche:
Can we get this merged?

Scalegj:
Please merge this, it works. I downloaded the key into my local repo and it fixed all my issues

poornack:
Confirming this is still needed on current `master` (314262c), plus some detail on why it only bites some setups.

**Symptom** — `./build-docker.sh` on an x86_64 host (Ubuntu 22.04, Docker 29.7, `i386/debian:trixie` image), armhf trixie build: stage0 fails with `E: Invalid Release signature`. `stage0/debootstrap.log`:

```
Signing key on A0DA38D0D76E8B5D638872819165938D90FDDD2E is not bound:
           No binding signature at time 2026-09-03T04:32:44Z
  because: Policy rejected non-revocation signature (PositiveCertification) requiring second pre-image resistance
  because: SHA1 is not considered secure since 2023-02-01T00:00:00Z
```

**Cause** — 67336fe re-signed the Raspberry Pi Archive key (82B129927FA3303E), but the Raspbian key (9165938D90FDDD2E) in `stage0/files/raspberrypi.gpg` still only carries its 2012 SHA-1 self-signatures. Per `gpg --list-packets`:

| Key | `master` | this PR |
|---|---|---|
| 82B129927FA3303E (Raspberry Pi Archive) | SHA-512 self-sigs, 2025-06-05 | same |
| 9165938D90FDDD2E (Raspbian) | SHA-1 self-sigs only, 2012-04-01 | adds SHA-256 self-sigs, 2025-06-12 (same as `raspbian-archive-keyring` 20120528.4) |

**Why Docker builds hit this and native builds don't** — trixie images ship `sqv` (apt 3.0 depends on it) but not `gpgv`, and debootstrap 1.0.141 only falls back to `sqv` when `gpgv` is absent. sqv's default policy rejects a key whose only binding signature is SHA-1; gpgv on a Raspberry Pi OS host accepts it, so the same stale keyring works natively.

**Test** — stock pi-gen trixie image with `gpgv` removed from PATH (sqv 1.3.0, debootstrap 1.0.141), against the live Raspbian trixie `InRelease` dated 2026-09-03 04:32:44 UTC:

| Keyring | `sqv --keyring … --cleartext InRelease` | `debootstrap --arch armhf --keyring … --print-debs trixie … http://raspbian.raspberrypi.com/raspbian/` |
|---|---|---|
| `master` | exit 1, error above | `E: Invalid Release signature` |
| this PR | exit 0 | `I: Valid Release signature`, resolves 184 packages |

The rootfs's own apt is unaffected either way, since it installs `raspbian-archive-keyring` 20120528.4, which already has the SHA-256 self-signature. Only the bootstrap copy in `stage0/files/` is stale.

XECDesign:
Sorry about the delay and thank you for the nudge.

poornack:
Thank you @XECDesign!

PR #866, opened 2025-08-28
Add cloud-init support

paulober:


paulober:
> Unless the user or rpi-imager provides a cloud-init configuration, it shouldn't do anything at all.
> 
> The idea isn't to replace pi-gen's configuration via cloud-init, but to allow users to make additional changes.

If that is what you want, then the best option would be to entirely remove cloud-init cfg file from the image > cloud-init wouldn't run. But then imager and others would need to have the config file and be able to store it in the system dir again before start configuring stuff.

PR #875, opened 2025-10-03
Drop powerstate change hook for rpi-usb-gadget

paulober:
Drop this reboot hook because the cloud-init `cc_raspberry_pi` module now includes a reboot trigger.

Related to: https://github.com/canonical/cloud-init/pull/6466

@tdewey-rpi

### DIFF
--- a/stage2/04-cloud-init/files/99_raspberry-pi.cfg
+++ b/stage2/04-cloud-init/files/99_raspberry-pi.cfg
@@ -5,12 +5,6 @@ datasource:
   NoCloud:
     seedfrom: file:///boot/firmware
 
-power_state:
-  delay: now
-  mode: reboot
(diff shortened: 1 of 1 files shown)


paulober:
The current approach checked after the first boot if there is the usb-gadget modprobe config present and if yes reboot. That results in boot to desktop and then reboot stuff. Now with some recommendations from canonical I moved the module to a different place in the cloud-init bootchain allowing it to directly execute the modifications. So now it can handle the reboot on its own after which cloud-init just continues with its boot process.

And coming back to your question, no it does not always reboot. It only reboots if the user requested via the cloud-init config to either enable the raspberry pi usb gadget or to enable one of the following interfaces: serial, i2c, spi or onewire.
(currently only for usb gadget and serial but i will update it for the others)

PR #885, opened 2025-11-06
Update README.md

laminarize:


laminarize:
updated verbiage from stating that ssh pub key entry should be a file. It should be a specific line item entry instead.

laminarize:
updated verbiage from stating that ssh pub key entry should be a file. It should be a specific line item entry instead.

PR #908, opened 2026-03-03
Unlock account when SSH public-key is set

stefanb2:
Commit 4b9cd15c6db9290d5e842503464e1d3429d92f37 breaks public-key SSH configuration.

Make sure first user has a shell when PUBKEY_SSH_FIRST_USER is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,9 +204,10 @@ The following environment variables are supported:
 
   * `PUBKEY_SSH_FIRST_USER` (Default: unset)
 
-   Setting this to a value will make that value the contents of the FIRST_USER_NAME's ~/.ssh/authorized_keys.  Obviously the value should
-   therefore be a valid authorized_keys file.  Note that this does not
-   automatically enable SSH.
+   Setting this to a value will make that value the contents of the FIRST_USER_NAME's
+   ~/.ssh/authorized_keys. Obviously the value should therefore be a valid authorized_keys file.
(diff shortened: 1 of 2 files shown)


XECDesign:
One of the first things build.sh enforces is that if you disable first user rename, you must set a password. If you ssh in while first boot wizard is running in the background, you'll end up in a bad situation. I can't find the issue report now, but that's why this commit was added in the first place - someone modified an image to enable ssh, added their key, sshed in, then ran though the wizard while logged in as the temporary first user.

Also, with a recent change that disables passwordless sudo by default, not setting a password can lock you out from doing anything useful on your system once you're in.

I'm not against the PR, but it needs to be done in a way that prevents such issues.

stefanb2:
I guess that makes sense from the Raspberry Pi Imager use case, but then `PUBKEY_SSH_FIRST_USER` & friends should be removed too, because they no longer work.

Let rework this.

stefanb2:
I realized that a new option isn't required at all and moved the code behind the `PUBKEY_SSH_FIRST_USER` conditional in stage2.

As `build.sh` only allows `FIRST_USER_PASS` to be unset when `DISABLE_FIRST_BOOT_USER_RENAME` is set to 0, your first concern should be addressed by this updated PR.

Passwordless sudo is a separate case. I.e. I'm aware that I have to enable that option during the next upstream rebase of my branches.

stefanb2:
Fixed the thinko in the PR.

Verified on HW that it restores SSH public key-only login on a password-less system.

PR #911, opened 2026-03-06
prerun.sh cleanup

itmaybeworks:
Always set `FAT_SIZE` to `32`, as `BOOT_SIZE` is statically set above and its unlikely that boot partition will be less than 128MB (`"134742016"`).


XECDesign:
I'd prefer it as-is, since there may be good reasons for a smaller boot partition.

PR #934, opened 2026-09-14
Add a practical guide for building custom images

klint-k:
This pull request adds a practical guide to using pi-gen.

I created the guide because I found that the existing documentation does not fully explain several important parts of the build process, particularly for someone creating and troubleshooting a custom Raspberry Pi OS image.

The guide covers:

* choosing the correct 32-bit or 64-bit branch;
* configuring a build;
* how stages and substages are processed;
* the exact order of package files, host scripts, and chroot scripts;
* creating a separate custom stage;
* using SKIP, SKIP_IMAGES, CLEAN, and CONTINUE;
* exporting the customized stage;
* incremental rebuilding;
* cloud-init and first-boot services;
* diagnosing package, chroot, mount, architecture, and storage failures;
* handling renamed or removed packages such as ntpdate.

The information was checked against the current pi-gen source code. The guide is intended to supplement the existing README with more detailed, task-oriented instructions.


XECDesign:
Thanks for the submission, but I can't accept this as-is. It's largely a restatement of the README, and a second 1000-line document isn't something I would like to maintain. If there's anything important the README doesn't make clear, I'd be happy to accept a more targeted PR fixing the existing documentation.

